PR #11, opened 2025-09-17
Add options to mute local alert sounds while the watch is connected

matejdro:
Not sure how the Core app does this functionality at the moment, maybe I have missed it. But, from what I could find, this seems to be the most elegant way of adding this feature: it mutes the spekaer/vibration from the phone, but it still forwards all notifications to the watch.

### DIFF
--- a/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
+++ b/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
@@ -11,12 +11,22 @@ import android.os.UserHandle
 import android.service.notification.NotificationListenerService
 import android.service.notification.StatusBarNotification
 import co.touchlab.kermit.Logger
+import io.rebble.libpebblecommon.LibPebbleConfigHolder
+import io.rebble.libpebblecommon.connection.ConnectedPebbleDevice
+import io.rebble.libpebblecommon.connection.LibPebble
+import io.rebble.libpebblecommon.connection.Watches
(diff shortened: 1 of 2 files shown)


libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
```diff
@@ -36,6 +48,10 @@ class LibPebbleNotificationListener : NotificationListenerService(), LibPebbleKo
     private val notificationHandler: NotificationHandler = get()
     private val connection: AndroidPebbleNotificationListenerConnection = get()
 
+    private val configHolder: LibPebbleConfigHolder = get()
```
sjp4: nit: you can use `NotificationConfigFlow` to avoid injecting the entire settings

matejdro: Changed it, but from what I can see, it's the exact same thing? The `flow` variable in the `NotificationConfigFlow` still exposes the entire settings.

sjp4: Oh that looks like an oversight. Reason behind adding a scoped holder for each settings category was just to make testing easier, without needing to provide all settings objects in a unit test

matejdro:
> Does muting phone calls work for you? I tried it out on my pixel, and muting notifications worked, but not calls

Just managed to test this, it works on my Pixel 9 Pro Fold. But it is a bit delayed, it does not appear to kick in immediately.

PR #14, opened 2025-10-02
Add option to respect phone DND

matejdro:
Small setting added to filter out all notifications that user muted with their DND/Work/Sleep/etc. modes.


ericmigi:
Just android right? any equivalent api on iOS?

PR #16, opened 2025-10-05
AppMessage classic PebbleKit Android implementation

matejdro:
PR adds support for the PebbleKit Android.

I recommend checking by commits: First PKJSLifecycleManager is generalized to allow non-JS companion apps, then basic AppMessage sending is implemented, then few bugs are fixed that prevented PebbleKit from functioning. Finally start/stop listeners and the provider is added.

To alleviate provider conflict issues somewhat, I have decided to not include the provider into the library's manifest. Instead a README entry is added to allow every implementing app to decide whether to use it or not.

I did not implement Data logging or sports/golf apps. I do not have any experience with this, so I have no way of testing whether these would work or not.

P.S.: I have noticed the message sending is quite a bit more sluggish than it used to be with the old Pebble app. Is that due to the use of the BLE vs BT Classic? EDIT:I've made a separate ticket for that.

### DIFF
--- a/libpebble3/src/commonTest/kotlin/io/rebble/libpebblecommon/connection/FakeAppMessages.kt
+++ b/libpebble3/src/commonTest/kotlin/io/rebble/libpebblecommon/connection/FakeAppMessages.kt
@@ -7,13 +7,13 @@ import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.flow.filter
+import kotlin.uuid.Uuid
 
 class FakeAppMessages: ConnectedPebble.AppMessages {
     val inbound = Channel<AppMessageData>(Channel.RENDEZVOUS)
     val outbound = Channel<AppMessageData>(Channel.RENDEZVOUS, onBufferOverflow = BufferOverflow.DROP_OLDEST)
     val outboundResults = Channel<AppMessageResult>(Channel.RENDEZVOUS, onBufferOverflow = BufferOverflow.DROP_OLDEST)
-
-    override val inboundAppMessages: Flow<AppMessageData> = inbound.consumeAsFlow()
 
     override suspend fun sendAppMessage(appMessageData: AppMessageData): AppMessageResult {
         return if (outbound.trySend(appMessageData).isSuccess) {
@@ -28,6 +28,8 @@ class FakeAppMessages: ConnectedPebble.AppMessages {
     override suspend fun sendAppMessageResult(appMessageResult: AppMessageResult) {
         outboundResults.send(appMessageResult)
     }
+
+    override fun inboundAppMessages(appUuid: Uuid): Flow<AppMessageData> = inbound.consumeAsFlow().filter { it.uuid == appUuid }
 
     override val transactionSequence: Iterator<UByte> = AppMessageTransactionSequence().iterator()
 }
--- a/libpebble3/src/commonTest/kotlin/io/rebble/libpebblecommon/js/PKJSRunnerTests.kt
+++ b/libpebble3/src/commonTest/kotlin/io/rebble/libpebblecommon/js/PKJSRunnerTests.kt
@@ -40,7 +40,7 @@ abstract class PKJSRunnerTests(
         appInfo: PbwAppInfo,
         lockerEntry: LockerEntry,
         jsPath: Path,
-        device: PebbleJSDevice,
+        device: CompanionAppDevice,
         urlOpenRequests: Channel<String>,
         logMessages: MutableSharedFlow<String>
     ) -> JsRunner
@@ -97,7 +97,7 @@ abstract class PKJSRunnerTests(
             APPINFO.copy(uuid = uuid.toString()),
             LOCKERENTRY.copy(id = uuid),
             writeJS(js),
-            PebbleJSDevice(
+            CompanionAppDevice(
                 watch.identifier,
                 watch.watchInfo,
                 appMessages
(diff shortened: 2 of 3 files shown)


matejdro:
Hm, not sure why build failed, it seems it failed in unrelated files. Any pointers?

matejdro:
No, this is an implementation of the old system (hence the "classic" in the name) to get compatibility with the old apps. I will start work on the new system soon, but it's more complex, so it will take a bit more.

ericmigi:
Ah my bad, missed that!

libpebble3/src/iosMain/kotlin/io/rebble/libpebblecommon/connection/endpointmanager/CompanionAppLifecycleManager.ios.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import io.rebble.libpebblecommon.connection.CompanionApp
+import io.rebble.libpebblecommon.js.CompanionAppDevice
+import io.rebble.libpebblecommon.metadata.pbw.appinfo.PbwAppInfo
+
+actual fun createPlatformSpecificCompanionAppControl(
+    device: CompanionAppDevice,
+    appInfo: PbwAppInfo
+): CompanionApp? {
+    // PebbleKit iOS is not supported (yet?)
```
sjp4: PebbleKit on iOS works directly through bluetooth without going through the Pebble app (i.e. this being a no-op is expected)

README.md
```diff
@@ -8,6 +8,20 @@ See https://github.com/coredevices/libpebble3/wiki/Roadmap
...
+
+```xml
+<provider
+    android:authorities="com.getpebble.android.provider.basalt"
+    android:name="io.rebble.libpebblecommon.pebblekit.PebbleKitProvider"
+    tools:ignore="ExportedContentProvider"
+    android:exported="true" />
+```
+
+Note that by doing so, your app will not be able to be installed alongside other apps that also do this.
```
sjp4, line 23: Probably worth noting that (unless we change libpebble so that it doesn't enable the gatt server by default) installing any other app using libpebble will break other apps (see `otherPebbleCompanionAppsInstalled`), so this might not really be a big problem. Leaving it like this is fine though

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/PebbleKitClassicStartListeners.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.util.UUID
+import kotlin.uuid.toKotlinUuid
+
+class PebbleKitClassicStartListeners() :
+    LibPebbleKoinComponent {
+    companion object {
+        private val logger = Logger.withTag(PebbleKitClassicStartListeners::class.simpleName!!)
+    }
+
+    private val context: Context = getKoin().get()
```
sjp4: This could probably be provided in `LibPebbleModule.android` and use constructor injection

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/PebbleKitClassicStartListeners.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val libPebble: LibPebble = getKoin().get()
+    private val connectionScope: LibPebbleCoroutineScope = getKoin().get()
```
sjp4: naming: this isn't a connection scope - it's a global scope

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/PebbleKitProviderNotifier.kt
```diff
@@ -0,0 +1,29 @@
...
+import io.rebble.libpebblecommon.di.LibPebbleKoinComponent
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
+
+class PebbleKitProviderNotifier : LibPebbleKoinComponent {
+    fun init() {
+        val watches: Watches = getKoin().get<LibPebble>()
+        val libPebbleCoroutineScope: LibPebbleCoroutineScope = getKoin().get()
+        val context: Context = getKoin().get()
```
sjp4: Same as above: could use constructor injection

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/endpointmanager/CompanionAppLifecycleManager.kt
```diff
@@ -0,0 +1,132 @@
...
+    companion object {
+        private val logger = Logger.withTag(CompanionAppLifecycleManager::class.simpleName!!)
+    }
+
+    private lateinit var device: CompanionAppDevice
+
+
+    private val runningApp: MutableStateFlow<CompanionApp?> = MutableStateFlow(null)
+    @Deprecated("Use more generic currentCompanionAppSession instead and cast if necessary")
+    override val currentPKJSSession: StateFlow<PKJSApp?> = PKJSStateFlow(runningApp)
```
sjp4, line 47: @crc-32 can you take a look as you're more familiar with the PKJS code (and AppMessage stuff)

README.md
matejdro: Hm, is the gatt server running even if no watches are set to connect?

sjp4: Correct (that's how the original Pebble app worked, so just copied that behaviour here). I think we found that it wasn't always reliable starting/stopping the GATT server on-demand 10 years ago; that might not be true now, though

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/endpointmanager/CompanionAppLifecycleManager.kt
crc-32: looks good!

crc-32:
@matejdro tested working with CoreApp after the above changes, please let me know if you disagree with any of them and if not the only outstanding things blocking this now are Steve's comments 🚀

PR #26, opened 2025-11-08
Initial support for the Pebblekit2

matejdro:
PR adds initial support for the https://github.com/pebble-dev/PebbleKitAndroid2 SDK. 

SDK is currently in the experimental phase, where we hope that some apps will try it and report any issues / missing functionality, so ideally, it should be supported in the Pebble apps, even in this phase.

I recommend checking by commits to break down this fairly large PR.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -19,7 +19,7 @@ kotlinx-datetime = "0.7.1"
 koin = "4.1.1"
 atomicfu = "0.29.0"
 compose-multiplatform = "1.9.3"
-pebblekit = "0.0.0"
+pebblekit = "510b3ead6e"
 settings = "1.3.0"
 kable = "0.40.0"
 kmpio = "0.2.3"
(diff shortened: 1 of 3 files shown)


matejdro:
It seems PebbleKit2 repo uses JDK 21 to build. Would you be opposed to me bumping `java-version` in the github action's `.yml˙ files?

sjp4:
> It has the same error than the #29 which was synced to develop yesterday. Was the CI fixed in develop since yesterday?

Yes, it was fixed last night (removing the github packages token requirement which broke CI)

gradle/libs.versions.toml
```diff
@@ -103,6 +104,9 @@ kotlinx-io-okio = { module = "org.jetbrains.kotlinx:kotlinx-io-okio", version.re
 koin-core = { group = "io.insert-koin", name = "koin-core", version.ref = "koin" }
 koin-compose = { group = "io.insert-koin", name = "koin-compose", version.ref = "koin" }
 koin-compose-viewmodel = { group = "io.insert-koin", name = "koin-compose-viewmodel", version.ref = "koin" }
+# Synthetic library declaration that then gets replaced with the included pebblekit, until
+# it is stable enough to make a release
+pebblekit = { module = "io.rebble.pebblekit2:server", version.ref = "pebblekit" }
```
sjp4: Could this use jitpack instead of a submodule? If you can easily replace it with e.g. `implementation("com.github.pebble-dev.PebbleKitAndroid2:server:2aaf338")` (jitpack is already enabled in settings.gradle) that would make it easier (because we'd have to also add the sobmodule to our private/internal repo and keep that in sync) - if not then that's OK

libpebble3/src/androidMain/AndroidManifest.xml
```diff
@@ -39,6 +39,8 @@
     <uses-feature android:name="android.hardware.bluetooth_le" android:required="false"/>
     <uses-feature android:name="android.software.companion_device_setup"/>
 
+    <uses-permission android:name="io.rebble.pebblekit2.permission.RECEIVE_DATA_FROM_WATCH" />
```
sjp4, line 42: Should this be defined in the pebblekit/server module?

sjp4 (approved):

gradle/libs.versions.toml
matejdro: That's a great idea, I totally forgot about Jitpack. Made the change.

libpebble3/src/androidMain/AndroidManifest.xml
matejdro: I'm a bit wary of implicit permissions. I prefer if the consumers explicitly declare all permissions they want to use.

PR #29, opened 2025-11-11
Improve reordering

### LINKED ISSUE #13: How does ordering of the system apps work?
I've started playing with the Locker now and I'm not sure how is user supposed to reorder the system apps on the watch.

`lockerApi.getLocker()` only returns user-installed apps + a single system app, calendar, which is not even on the menu, so reordering it does not make sense. 

Or is that something that has not been implemented yet?

matejdro:
PR makes it possible to reorder all apps in the watch's menu. It makes two changes:

1. It inserts system apps into locker db. This is needed, because system apps also need their `orderIndex` to make ordering work. I would appreciate some feedback here, especially on the way I blocked system apps from being synced via blob db. I have also removed Calendar from the `SystemApps` and added all apps that appear in the menu on my P2 (I can add more system apps if there are more on the P2D), so this might cause issues with the way Core app does calendar settings. I'm also not sure what implications this has on the locker sync, if any (microPebble has no locker sync, so I cannot test that).
2. It fixes the  `lockerApi.setAppOrder()` to properly set the order indexes. When the app is moved to the desired spot, it will also fix order indexes of all in-between apps, so the numbers will be valid after the operation.

This fixes #13


blobdbgen/src/main/kotlin/coredev/BlobDbEntityProcessor.kt
```diff
@@ -389,6 +389,13 @@ class BlobDbEntityProcessor(
                         """.trimIndent())
                     .build()
             )
+
+            val disableSyncOnSystemEntries = if (annotation.databaseId == BlobDatabase.App) {
```
sjp4: I think this value (setting sync to false) will get over-written when `updateSync()` is called (because we set the value of sync based on the `orderIndex`).

If we were going to have custom logic here, i'd like. it to be controlled by the annotation (rather than this code knowing about specific database IDs) - but hopefully we can avoid changes in the processer itself.

Can you do this in `updateSync` in `LockerEntryRealDao`? (that could filter on whether it's a system app..)

The only other thing is that some system apps are only applicable to some types of watch/fw versions. I don't see any easy way to handle that with the generic way we process blobdb sync right now - but it hopefully won't break anything if we leave in a system app which isn't on a given watch.

matejdro: Ah it seems I have only changed `updateOrder`:

https://github.com/coredevices/libpebble3/pull/29/files#diff-d1ab3249d19284998474aa1758b8fbd0d816b8de17a6b01fa84c13eb51a207afR130

Will also do `updateSync`

sjp4: I'm going to think a bit more on how we can handle the sync limit - as it stands, if we have a sync limit of 20 and 8 system apps, then only 12 non-system apps will be inserted. And if system apps are moved outside the sync limit (e.g. position 21) then they won't be on the list sent by the app order endpoint.

sjp4 (approved):

matejdro:
In case you are interested, this is how we solved the DB out-of-sync issue: https://github.com/matejdro/libpebble3/commit/b9d14983eeab9c439c2cbf3368db86dd3cc36a14

PR #50, opened 2025-11-30
Add custom vibration patterns for notifications

FlynnD273:
This is just changing the backend, and adding a custom vibration pattern to the "Write Notification" button in the app. Currently uses UInt32 lists.

### DIFF
--- a/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotification.kt
+++ b/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotification.kt
@@ -30,6 +30,7 @@ data class LibPebbleNotification(
     val icon: TimelineIcon,
     val actions: List<LibPebbleNotificationAction>,
     val people: List<String>,
+    val vibrationPattern: List<UInt>,
     val color: Int? = null, // ARGB
 ) {
     fun displayDataEquals(other: LibPebbleNotification): Boolean {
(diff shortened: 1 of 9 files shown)


libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationEntity.kt
```diff
@@ -24,4 +23,6 @@ data class NotificationEntity(
     val decision: NotificationDecision,
     @ColumnInfo(defaultValue = "NULL")
     val people: List<String>? = emptyList(),
+    @ColumnInfo(defaultValue = "")
```
sjp4: I don't think it's needed in this table (which is only used for lookups on previous notifications e.g. for duplicates/UI)

libpebble3/schema/io.rebble.libpebblecommon.database.Database/24.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 24,
-    "identityHash": "4debf03d994cda016c2cb7b40242c919",
```
sjp4, line 5: If you are adding a field then the database version should be bumped to 25 (migration 24 has already gone out) - but I don't think you need that if you remove the change to `NotificationEntity`

sjp4:
Very cool change BTW - this is one of the things I really want! (I will build out some UI in the app to set patterns per-app/contact/etc)

FlynnD273:
Sounds good! I only added the database change because it looked like other notification fields were there too. If I don't need to change that then I won't :shrug:

FlynnD273:
A stretch goal would be to somehow allow Tasker to send custom vibration patterns to the watch. My specific use-case is that I don't have notifications shown on my watch, I just set up per-app and per-person within each app vibration patterns. But this is a more niche use and I suspect I may have to end up maintaining my own fork if I want that kind of behaviour.

PR #54, opened 2025-12-06
health integration

michaelthatsit:
this is pretty early. all it does is pull the health data from the watch, and push the averages to it.

I'm thinking it's probably best to add this and do any UI changes in a separate PR: In it I would probably add:

TODOs:

- [x] set averages on watch
- [x] sync multiple watches

### DIFF
--- a/experimental/src/commonMain/kotlin/coredevices/ExperimentalDevices.kt
+++ b/experimental/src/commonMain/kotlin/coredevices/ExperimentalDevices.kt
@@ -7,6 +7,7 @@ import CoreRoute
 import DocumentAttachment
 import coredevices.pebble.ui.TopBarParams
 import coredevices.util.Permission
+import kotlinx.io.files.Path
 
 class ExperimentalDevices {
     fun init() {


Odel:
Would this handle a user switching between two watches throughout the day?

michaelthatsit:
Not yet no, and some of the discussion on discord suggests some of my initial assumptions were incorrect on how the data storage works on the watch.

This solution works but some steps I'm taking are unnecessary and could impact battery life.

michaelthatsit:
May have gone a bit overboard here. But here's the daily, weekly, and monthly screens for health data. Monthly needs work and as of now you can only view the current day, week, or month. 

Monthly is meant to show an area graph of each week of the current month but I'm still playing with KoalaPlot. 

Screenshot_20251216-195228
Screenshot_20251216-195242
Screenshot_20251216-195253

michaelthatsit:
> It might be relevant to have the Unit Distance (Miles/Kilometers) setting in the health integration of the mobileapp, if it is not already the case in this PR

I added this in the health settings panel but it doesn't seem to immediately update on the watch.

michaelthatsit:
Also units work locally but it doesn't update the watch.

michaelthatsit:
> Looks nice, but it would be great if there where an scale on the Y-axis.

Ah I'm still debating this. if you consider the ethos of the device, it's not meant to be a detailed fitness tracker like fitbit, AW, Garmin. This functionality is more for tracking trends and vibes. 

We care that "number go up," not what that number actually is.

Mr-Bajs:
Sure, but I don't think a simple discrete scale goes against the ethos. I would personally prefer it. It sorts of reminds you of your average between long periods.

asyba:
If adding a new tab is not possible, maybe the Health statistics could be inside in the "Apps" tab, but not as an app that we must click, maybe a section above the apps list.

michaelthatsit:
A really rough sketch of what I mean. 

Watches could be swapped using a drop down or modal.

michaelthatsit:
@sjp4 Finished the separation of UI and did a little clean up. This PR now only contains watch syncing logic.

I've tested it and everything works as expected, it's now ready for your review.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt
```diff
@@ -100,21 +118,27 @@ sealed class UserFacingError {
     data class FailedToSideloadApp(override val message: String) : UserFacingError()
     data class FailedToScan(override val message: String) : UserFacingError()
     data class MissingCompanionApp(
-        override val message: String,
-        val appName: String,
-        val appUuid: Uuid,
-        val downloadUrl: String?
+            override val message: String,
```
sjp4: I think you have some different code formatting settings than the android studio defaults? A lot of the formatting changes here seem too indented (either way - probably best not to include formatting changes to unchanged code)

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt
```diff
@@ -100,21 +118,27 @@ sealed class UserFacingError {
...
-    val name: String,
+        val pkg: String,
+        val name: String,
 )
 
 interface HealthApi {
     val healthSettings: Flow<HealthSettings>
+    val healthUpdateFlow: Flow<Unit>
     fun updateHealthSettings(healthSettings: HealthSettings)
+    suspend fun getHealthDebugStats(): io.rebble.libpebblecommon.services.HealthDebugStats
```
sjp4: nit: should be able to import this, rather than fully-qualified reference

pebble/src/commonMain/kotlin/coredevices/pebble/watchModule.kt
```diff
@@ -150,6 +152,11 @@ val watchModule = module {
         )
     }
 
+    // Provide HealthDao from LibPebble's internal Koin instance
+    single {
```
sjp4: Don't do this :) (everything should go via the `LibPebble` API - one day that will be enforced in a way where the internals won't be visible from here). I don't think you're actually using this anywhere in the app code anyway?

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/di/LibPebbleModule.kt
```diff
@@ -392,8 +401,10 @@ fun initKoin(
                 singleOf(::MissedCallSyncer)
                 singleOf(::FirmwareDownloader)
                 singleOf(::JsTokenUtil)
-                singleOf(::Datalogging)
-                singleOf(::Health)
+                single { Datalogging(get(), get(), get(), get()) }
```
sjp4: Should be able to continue to use the `singleOf(::XX)` syntax for these

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/di/LibPebbleModule.kt
```diff
@@ -282,6 +285,11 @@ internal interface LibPebbleKoinComponent : KoinComponent {
     override fun getKoin(): Koin = LibPebbleKoinContext.koin
 }
 
+/**
+ * Public API to access LibPebble's internal Koin instance for getting DAOs and other internal services
+ */
+fun getLibPebbleKoin(): Koin = LibPebbleKoinContext.koin
```
sjp4: Shouldn't be needed (see other comment)

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/datalogging/Datalogging.kt
```diff
@@ -3,43 +3,344 @@ package io.rebble.libpebblecommon.datalogging
...
+                MEMFAULT_CHUNKS_TAG -> {
+                    libPebbleCoroutineScope.launch {
+                        val chunk = MemfaultChunk()
+                        chunk.fromBytes(DataBuffer(data.toUByteArray()))
+                        val chunkBytes = chunk.bytes.get()
+                        webServices.uploadMemfaultChunk(chunkBytes.toByteArray(), watchInfo)
+                    }
+                }
+                HEALTH_STEPS_TAG -> {
+                    libPebbleCoroutineScope.launch {
```
sjp4: Probably should move the parsing code out of this class into a health-specific class

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/datalogging/Datalogging.kt
```diff
@@ -3,43 +3,344 @@ package io.rebble.libpebblecommon.datalogging
...
 import kotlin.uuid.Uuid
+import kotlinx.coroutines.launch
 
 class Datalogging(
-    private val libPebbleCoroutineScope: LibPebbleCoroutineScope,
-    private val webServices: WebServices,
+        private val libPebbleCoroutineScope: LibPebbleCoroutineScope,
+        private val webServices: WebServices,
+        private val healthDao: HealthDao,
+        private val knownWatchDao: io.rebble.libpebblecommon.database.dao.KnownWatchDao,
```
sjp4: import

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt
michaelthatsit: Yeah I was trying out antigravity while doing a final pass and it the model just reorganizes code for some reason.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/dao/HealthDao.kt
```diff
@@ -0,0 +1,151 @@
...
+import androidx.room.Dao
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import io.rebble.libpebblecommon.database.entity.HealthDataEntity
+import io.rebble.libpebblecommon.database.entity.OverlayDataEntity
+import kotlinx.coroutines.flow.Flow
+
+@Dao
+interface HealthDao {
```
sjp4, line 12: This needs a cleanup method to delete expired data (and so a policy for when data is expired). That should be called from `Housekeeping` to ensure it happens regularly

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/datalogging/Datalogging.kt
```diff
@@ -16,30 +24,135 @@ import kotlin.uuid.Uuid
...
+                        }
+
+                        // Parse the health steps data using the centralized parser
+                        val records = parseStepsData(data, itemSize)
+                        if (records.isEmpty()) {
+                            logger.w { "No health step records parsed" }
+                            return@launch
+                        }
+
+                        logger.i {
```
sjp4: Default log level should be `d` unless something unusual is happening

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/health/Health.kt
```diff
@@ -1,33 +1,272 @@
...
+
+    fun isActive(service: HealthService): Boolean =
+            synchronized(lock) { activeService == service }
+}
 
 class Health(
-    private val watchSettingsDao: WatchSettingsDao,
-    private val libPebbleCoroutineScope: LibPebbleCoroutineScope,
-): HealthApi {
+        private val watchSettingsDao: WatchSettingsDao,
```
sjp4: formatting (all the new code still has class member `val`s indented too far)

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/health/Health.kt
```diff
@@ -1,33 +1,272 @@
...
+    fun forceSyncLast24Hours()
+}
+
+/**
+ * Implementation that finds connected pebbles and calls their HealthService
+ *
+ * This uses a registry pattern where HealthService instances register themselves when they're
+ * created for a connection.
+ */
+class RealHealthServiceAccessor(
```
sjp4: I'm don't think `RealHealthServiceAccessor`/`HealthServiceRegistry` is needed - callers should be able to iterate through connected watches (see e.g. the `sendPing` / `launchApp` methods in `Libpebble3`). That would invoke a method on `ConnectedPebbleDevice` where we would add a new `Health` interface (which could be implemented by the health service).

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthService.kt
```diff
@@ -0,0 +1,618 @@
...
+            MutableStateFlow(
+                    true
+            ) // When false, drops all incoming health data (used during reconciliation)
+
+    companion object {
+        private val logger = Logger.withTag("HealthService")
+
+        private val HEALTH_TAGS =
+                setOf(HEALTH_STEPS_TAG, HEALTH_SLEEP_TAG, HEALTH_OVERLAY_TAG, HEALTH_HR_TAG)
+        private const val HEALTH_STEPS_TAG: UInt = 81u
```
sjp4: Looks like these are also defined in `Datalogging` - ideally define in one place and use from the other

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthService.kt
```diff
@@ -0,0 +1,618 @@
...
+        private const val HEALTH_SLEEP_TAG: UInt = 83u
+        private const val HEALTH_OVERLAY_TAG: UInt = 84u
+        private const val HEALTH_HR_TAG: UInt = 85u
+
+        private const val TWENTY_FOUR_HOURS_MS = 24 * 60 * 60_000L
```
sjp4: `1.days.inWholeMilllisecnods` will do the same thing

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthService.kt
```diff
@@ -0,0 +1,618 @@
...
+        private const val HEALTH_STATS_AVERAGE_DAYS = 30
+        private const val HEALTH_SYNC_WAIT_MS = 8_000L
```
sjp4: could use 8.seconds.inWholeMilliseconds`

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthService.kt
```diff
@@ -0,0 +1,618 @@
...
+                // Update stats with the new data
+                updateHealthStats()
+                lastFullStatsUpdate.value = System.now().toEpochMilliseconds()
+            } else {
+                logger.i { "HEALTH_SERVICE: Force sync completed - no new data from watch" }
+            }
+        }
+    }
+
+    private fun isActiveConnection(reason: String): Boolean {
```
sjp4: FWIW I wouldn't worry too much about trying to solve the multiple-connected-watches case right now - that is something we have to solve more generally (probably including selecting one watch as "primary" or similar). I would probably not make that a concern of the service - it just does what it is told for a given watch

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthStatsSync.kt
```diff
@@ -0,0 +1,428 @@
...
+
+/** Sends a health stat to watch via BlobDB */
+private suspend fun sendHealthStat(
+        blobDBService: BlobDBService,
+        key: String,
+        payload: UByteArray
+): Boolean {
+    val response =
+            withTimeoutOrNull(HEALTH_STATS_BLOB_TIMEOUT_MS) {
+                blobDBService.send(
```
sjp4: The health service shouldn't be writing BlobDb protocol messages directly - we should create a new database table for `HealthStats` using `@GenerateRoomEntity` (and add it to `BlobDbDaos`), then any records inserted into that table will be automatically synced to every watch (even if not connected at the time of writing).

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthService.kt
```diff
@@ -0,0 +1,618 @@
...
+                .distinctUntilChanged { old, new -> old == new }
+                .onEach { isAppOpen.value = it }
+                .launchIn(scope)
+
+        protocolHandler
+                .inboundMessages
+                .onEach { packet ->
+                    when (packet) {
+                        is DataLoggingIncomingPacket.OpenSession -> handleSessionOpen(packet)
+                        is DataLoggingIncomingPacket.SendDataItems -> handleSendDataItems(packet)
```
sjp4: Is this handling all duplicated from `Datalogging`?

michaelthatsit:
The refinement continues, I went ahead and consolidated all datalogging into DataLoggingService. Ready for re-review.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthService.kt
michaelthatsit: it is. I went ahead and removed this, consolidating all DataLogging into DataLoggingService.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/services/HealthStatsSync.kt
michaelthatsit: Done!

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/Housekeeping.kt
```diff
@@ -29,5 +31,12 @@ class Housekeeping(
     private suspend fun doHousekeeping() {
         val deleteNotificationsOlderThan = clock.now() - notificationConfigFlow.value.storeNotifiationsForDays.days
         notificationsDao.deleteOldNotifications(deleteNotificationsOlderThan.toEpochMilliseconds())
+
+        val deleteHealthDataOlderThan = clock.now() - 90.days
+        val deletedHealthRecords = healthDao.deleteExpiredHealthData(deleteHealthDataOlderThan.epochSeconds)
+        val deletedOverlayRecords = healthDao.deleteExpiredOverlayData(deleteHealthDataOlderThan.epochSeconds)
+        if (deletedHealthRecords > 0 || deletedOverlayRecords > 0) {
+            println("Housekeeping: Deleted $deletedHealthRecords health records and $deletedOverlayRecords overlay records older than 90 days")
```
sjp4: use `Logger` instead of `println`

asyba:
@michaelthatsit Are these new tables designed for future synchronization with Apple Health or Google Health?
We don't want to run into problems later when we want to implement it, making us say, "We should have done this from the beginning," and now we have to do some workarounds, etc.

michaelthatsit:
> @michaelthatsit Are these new tables designed for future synchronization with Apple Health or Google Health?
> 
> We don't want to run into problems later when we want to implement it, making us say, "We should have done this from the beginning," and now we have to do some workarounds, etc.

I do believe they're at this point compatible but I can double check. There's a library I mentioned in an [open issue](https://github.com/coredevices/mobileapp/issues/76) that we can use for quick integrations for both services. 

If the plan is to simply integrate with those services and not provide a built in health panel, we should also probably not store more than a week of data at a time.

asyba:
I vote to have them on boths places on pebble app and then just send them on apple/google/etc in the background.

sjp4:
I pushed a bunch of fixes - see the commit message for details. A few things were broken which was stopping me from testing anything, and noticed more (including refactoring the existing settings model) as I went. I think sync requests and updating settings are working now - can you try out the updated branch, too? I'll go through the parsing/state code later - would be good to know it everything is still working for you before merging though - I made quite a lot of changes.

PR #62, opened 2025-12-13
Introduce concept of app quirks

KonradIT:
Noticed the twitter app sends the DMs with repeated subject name in the title. Format is: "Sender: Sender", so someone with a really long name occupies the entire title and doesn't look good.

I have other quirks in mind but wanted to get feedback on how to better implement this.


libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
```diff
@@ -50,7 +50,8 @@ class BasicNotificationProcessor(
             channel,
             notificationConfigFlow.value
         )
-        val title = sbn.notification.extras.getCharSequence(Notification.EXTRA_TITLE) ?: ""
+
+        val title = NotificationProperties.lookup(app.packageName)?.title?.invoke(sbn, app, channel) ?: sbn.notification.extras.getCharSequence(Notification.EXTRA_TITLE) ?: "",
```
sjp4: nit: there's a `val appProperties` created below this, which you could move further up and re-use, to avoid looking up the package name twice.

sjp4:
FYI CI should be working now - looks like some real build errors

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/NotificationProperties.kt
```diff
@@ -3,76 +3,87 @@ package io.rebble.libpebblecommon.notification.processor
...
 import io.rebble.libpebblecommon.timeline.TimelineColor
 
+fun title(sbn: StatusBarNotification, notiApp: NotificationAppItem, channel: ChannelItem?): String {
+    return sbn.notification.extras.getCharSequence(Notification.EXTRA_TITLE) as String
+}
 enum class NotificationProperties(
     val pkgName: String,
     val color: TimelineColor?,
     val icon: TimelineIcon?,
+    val title: (StatusBarNotification, NotificationAppItem, ChannelItem?) -> String,
```
matejdro, line 13: Maybe we could set the default here, so we don't have to specify it for all apps explicitly?

PR #65, opened 2025-12-15
Enable sort notification list for iOS

asyba:
Added sorting dropdown for the Notification list section for iOS:

Android: Uses pebbleFeatures.supportsNotificationAppSorting() which returns true only for Android
This displays a filter chip row below the search bar with two chips: "Notified only" and a sort dropdown.
This adds an entire extra row to the UI.

iOS: I added a sort icon button directly in the toolbar actions, next to the search bar.
This approach doesn't add an extra row/bar to the UI.
I couldn't reuse supportsNotificationAppSorting() because it would show the "Notified only" that is specific for Android.
 
The "Count" sorting option is now hidden on iOS since doesn't provide access to notification history like Android does.


asyba:
> The goal with `PebbleFeatures` is to remove and `isIos()`/`isAndroid()` checks from the UI code, so ideally we should just add more functions to `PebbleFeatures` to support making this change. I think ideally we'd keep the UI consistent between the platforms, too (i.e. using filter chips, rather than a dropdown in the `topBar`. We could probably do this without much code change by adding another `PebbleFeatures` entry or two, and a couple more conditionals in the existing filter row code.

got it.
what should be done for the "Notified Only" since for iOS is not supported.
I try removing but the UI shows in the middle and bad.
maybe show the "Notified Only" but as disabled? or I have to move the "Recent" to the right size?
image

PR #73, opened 2025-12-20
Respect vibration patterns from Notification Channel and Tasker

FlynnD273:
Android doesn't really do per-notification vibration patterns anymore as far as I can tell. This PR pulls the vibration pattern from the Notification Channel, and from the extra field in Tasker's AutoNotification plugin. The Notification Channel code adds a 0 at the beginning of the pattern to match Android's vibration pattern syntax, and the Autonotification code doesn't, to match with Pebble's vibration pattern syntax. This replaces #71.

### DIFF
--- a/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
+++ b/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
@@ -127,7 +127,7 @@ class LibPebbleNotificationListener : NotificationListenerService(), LibPebbleKo
                     id = channel.id,
                     name = channel.name.toString(),
                     muteState = MuteState.Never,
-                    vibrationPattern = channel?.getVibrationPattern()?.map { it.toUInt() }?.let { listOf(0u) + it }
+                    vibrationPattern = channel.getVibrationPattern().map { it.toUInt() }.let { listOf(0u) + it }
                 )
                 val group = groups[channel.group]
                 if (group == null) {
--- a/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
+++ b/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
@@ -115,7 +115,6 @@ class BasicNotificationProcessor(
         sbn: StatusBarNotification,
         channel: ChannelItem?,
     ): List<UInt>? {
-        logger.w { "vibration pattern info: ${sbn.getNotification().extras.getString("extraautonotificationinfo", "")}" }
         // TODO we're only picking the pattern from the first contact. I don't know if the first
         //  contact is always the one that sent the message, in a group chat?
         val vibePatternForContact = findVibePattern(contactEntries.firstOrNull()?.vibePatternName)


libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
```diff
@@ -113,18 +113,32 @@ class BasicNotificationProcessor(
         contactEntries: List<ContactEntity>,
         app: NotificationAppItem,
         sbn: StatusBarNotification,
+        channel: ChannelItem?,
     ): List<UInt>? {
+        logger.w { "vibration pattern info: ${sbn.getNotification().extras.getString("extraautonotificationinfo", "")}" }
```
sjp4: Logging this should go in `StatusBarNotification.dump()` in `NotificationHandler.kt`, rather than here

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
```diff
@@ -127,6 +127,7 @@ class LibPebbleNotificationListener : NotificationListenerService(), LibPebbleKo
                     id = channel.id,
                     name = channel.name.toString(),
                     muteState = MuteState.Never,
+                    vibrationPattern = channel?.getVibrationPattern()?.map { it.toUInt() }?.let { listOf(0u) + it }
```
sjp4: Does `getVibrationPattern()` always return something? (i.e. will the field always be non-null?) Documentation doesn't say it's nullable..

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
```diff
@@ -113,18 +113,32 @@ class BasicNotificationProcessor(
...
+        } else {
+            null
+        }
         val vibePatternFromNotification = if (notificationConfigFlow.value.useAndroidVibePatterns) {
             sbn.notification.vibrationPattern()
         } else {
             null
         }
+        val vibePatternFromTaskerNotification = if (notificationConfigFlow.value.useAndroidVibePatterns) {
+            var patt = emptyList<UInt>()
```
sjp4, line 133: Probably would do something like:

```
val vibePatternFromTaskerNotification = if (notificationConfigFlow.value.useAndroidVibePatterns) {
            try {
                sbn.notification.extras.getString("extraautonotificationinfo", "").split(",").map { it.toUInt() }.ifEmpty { null }
            } catch (e: Exception) {
                logger.w(e) { "Error parsing vibe pattern" }
                null
            }
        } else {
            null
        }
```

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
```diff
@@ -111,6 +111,7 @@ data class ChannelItem(
     val id: String,
     val name: String,
     val muteState: MuteState,
+    val vibrationPattern: List<UInt>? = null,
```
sjp4, line 114: This won't work unless you bump the database version and add a migration to create the column - see e.g. https://github.com/coredevices/CoreApp/commit/d1e34ef7116aa69170ff47ba9561a144a8bd9873#diff-4bbbd39fe1ff5faa4ceef66355eb272f66db7ac0e0e462a95647c1b6c4272ea7 (otherwiswe it will crash, unless you're running a fresh install of the app)

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
FlynnD273: Oops I meant to remove that line entirely

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
FlynnD273: Unsure. But channel is nullable.

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
FlynnD273: Should it log those? The only error you'll get is a numberformatexception, and the most likely cause of that will be because the extra info contains info unrelated to the vibration pattern.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
FlynnD273: Huh. It was crashing before I added the default value, but it was working fine for me after I set the default value. Maybe it just ignores that column in the db right now?

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
sjp4: Ah right - actually there `channel` isn't nullable (you could remove all the `?` in that line and it would compile - it's smartcasted to non-null. That essentially means that we'll never store a null value here (so the code below that reads it will never get past `vibePatternForChannel` in the hierarchy of patterns it checks for). Might be worth checking what (if anything) is put here by default, and testing for that

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
sjp4: Whatever you think makes sense - you've got a better idea of the context of what would be in that extra

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
sjp4: Oh my bad - got confused with db fields vs this object which is serialized into a db field (so adding the default value fixing it makes sense)

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotificationListener.kt
FlynnD273: Oops yeah I was confusing it with the BasicNotificationProcessor code. I don't think channel could ever be null there, ever. Like, channel is always `NotificationChannel` and not `NotificationChannel?`

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
FlynnD273: Sounds good. So I think I can just leave it, right? The vibration pattern just won't get added to the db as a column?

PR #82, opened 2025-12-27
Fix timeout issue sideload FW on iOS

asyba:
There's a problem on iOS with uploading custom FW, it almost always fails due to a timeout or because very large data packets are sent in a short period, causing the BL connection to fail.

Lowering the desiredRxWindow value from the default 25 to 6 has improved things. It worked well for me about 95% of the time without any issues.

However, this change affects not only sideloading but all communication between the Pebble and iPhone.
I notice that watchfaces/apps upload a little slower than usual with 25, notifications works as normal as before.

Perhaps we should look into a way to isolate the use of desiredRxWindow=6 for sideloading only, but that would require significant changes.
Old pebble supposedly use desiredRxWindow with 4.

I'm not sure if the same thing is happening on Android, or if it works fine and the problem is on iOS where large packets in a short period cause the BL connection to collapse.

### DIFF
--- a/libpebble3/src/iosMain/kotlin/io/rebble/libpebblecommon/connection/bt/ble/transport/GattServer.ios.kt
+++ b/libpebble3/src/iosMain/kotlin/io/rebble/libpebblecommon/connection/bt/ble/transport/GattServer.ios.kt
@@ -226,6 +226,7 @@ actual class GattServer(
         return try {
             withTimeout(SEND_TIMEOUT) {
                 while (true) {
+                    val startCount = peripheralManagerReadyCount.value
                     if (peripheralManager.updateValue(
                             value = data.toNSData(),
                             forCharacteristic = cbCharacteristic,
@@ -235,7 +236,7 @@ actual class GattServer(
                         return@withTimeout true
                     }
                     // Write did not succeed; wait for queue to drain
-                    peripheralManagerReady.first()
+                    peripheralManagerReadyCount.first { it > startCount }
                 }
                 false
             }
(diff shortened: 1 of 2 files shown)


libpebble3/src/iosMain/kotlin/io/rebble/libpebblecommon/connection/bt/ble/transport/GattServer.ios.kt
```diff
@@ -235,7 +236,7 @@ actual class GattServer(
                         return@withTimeout true
                     }
                     // Write did not succeed; wait for queue to drain
-                    peripheralManagerReady.first()
+                    peripheralManagerReadyCount.first { it > startCount }
```
sjp4, line 239: This looks like a separate change - did you observe a problem (or just see a problem that you thought needed fixing?) Either way, this as a separate PR would be great, rather than bundled with the other change

sjp4:
> There's a problem on iOS with uploading custom FW, it almost always fails due to a timeout or because very large data packets are sent in a short period, causing the BL connection to fail.

Did you see the same thing with app installs, or only installing firmware? They use the same putbytes mechanism. I'm also curious if you submitted a bug report using the app (because i'd be interested in seeing the watch logs from when this happens).

> Perhaps we should look into a way to isolate the use of desiredRxWindow=6 for sideloading only, but that would require significant changes.

I think that's unlikely to be feasible - without re-establishing the connection each time we want to switch. 

> Old pebble supposedly use desiredRxWindow with 4.

That sounds believable. 6 sounds reasonable. Ideally I think we'd make this configurable rather than hardcoded, but that would be in `BleConfig` and we can't vary the default values per-platform easily, there (we don't want to make it 6 on android). This solutions sounds OK to me.

libpebble3/src/iosMain/kotlin/io/rebble/libpebblecommon/connection/bt/ble/transport/GattServer.ios.kt
asyba: At first I did this change it, can't remember if improved a little or not.
since then it was still happening the issue, I changed the desiredRxWindow.

Maybe the desiredRxWindow is the only part that will fix it.
I could test only with this change and remove the others.

asyba:
@sjp4 I just did a bug report check #8745

sjp4:
We also saw some issues internally lately - I made some changes which may help

Issue #81, opened 2025-12-26
ConnectivityStatus out of bounds


matejdro:
Any ideas what that could be caused by?

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
	at io.rebble.libpebblecommon.connection.bt.ble.pebble.ConnectivityStatus.<init>(ConnectivityWatcher.kt:91)
	at io.rebble.libpebblecommon.connection.bt.ble.pebble.ConnectivityWatcher$subscribe$2.invokeSuspend(ConnectivityWatcher.kt:57)
```

I have managed to trigger it when pairing to a watch. This is with the 8a4a394b4ece73f2b1d6de7cd69fe5104bbce413 lib version.

PR #83, opened 2025-12-28
Fix crash when connecting to watches in recovery firmware

g150446:
## Problem

When connecting to a Pebble watch running **recovery firmware** (PRF), the app crashes with an `ArrayIndexOutOfBoundsException` in `ConnectivityWatcher.kt:91`.

**Root Cause:**
- Recovery firmware returns an **empty byte array** (length=0) for the BLE connectivity characteristic
- The `ConnectivityStatus` constructor attempts to access `characteristicValue[0]` and `characteristicValue[3]` without checking if the array is empty
- This causes an immediate crash during connection, preventing users from updating their watch firmware

**User Impact:**
- Users cannot connect to watches in recovery mode
- Firmware updates cannot be completed
- App shows "Rebooting after update - Waiting" indefinitely after firmware update

**Error from logs:**
```
java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
    at io.rebble.libpebblecommon.connection.bt.ble.pebble.ConnectivityStatus.<init>(ConnectivityWatcher.kt:91)
```

## Solution

Added bounds checking in the `ConnectivityStatus` constructor to handle empty byte arrays from recovery firmware:

1. **Check for empty array** - If `characteristicValue.isEmpty()`, initialize all fields with safe default values (all false, NO_ERROR)
2. **Add bounds check for pairingErrorCode** - Only access `characteristicValue[3]` if array size > 3

**Files Changed:**
- `libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/bt/ble/pebble/ConnectivityWatcher.kt`

## Testing

- Tested on Pixel 7a (Android 14) with Pebble Time in recovery mode (v3.0.1-prf-rc1)
- Successfully connects to recovery firmware
- Firmware update to v4.4.3-rbl completes successfully
- Watch reconnects properly after firmware update

### DIFF
--- a/pebble/src/commonMain/kotlin/coredevices/pebble/ui/DebugFirmwareSideload.kt
+++ b/pebble/src/commonMain/kotlin/coredevices/pebble/ui/DebugFirmwareSideload.kt
@@ -84,7 +84,15 @@ private sealed class UiFirmwareUpdateStatus {
                     "Couldn't start: ${status.error.message()}"
                 )
 
-                is FirmwareUpdater.FirmwareUpdateStatus.NotInProgress.Idle -> Idle
+                is FirmwareUpdater.FirmwareUpdateStatus.NotInProgress.Idle -> {
+                    // Check if there was a failure
+                    val failure = status.lastFailure
+                    if (failure != null) {
+                        Error(failure, failure.message ?: "Unknown error")
+                    } else {
+                        Idle
+                    }
+                }
             }
         }
     }


sjp4:
(related: https://github.com/coredevices/mobileapp/issues/81)

Were you able to reproduce this regularly, to show that recovery FW always returns an empty characteristic value for this? We had been this happen a couple of times but it seemed random (i.e. like some kind of FW corruption bug, perhaps) - I would have assumed that we would have had a lot more stuck users if it really happened every time a Time watch was in PRF.

Regardless, the fix it probably the same - but e.g. if it was just a transient thing which was fixed upon reconnection then i'd be tempted to leave it as-is (rather than assume the device is not paired and start repairing). But I have never reproduced it so your input is welcome...

sjp4:
Ping @g150446 re: question above

PR #84, opened 2025-12-29
Remove Rebble login requirement for the Rebble app feed

FlynnD273:
It still seems to disable all Rebble categories by default but I couldn't figure out how to fix that


sjp4:
Right now, users have to be logged in to have a good experience using the rebble feed (there's various bits of metadata are only in the locker API, not store API, which libpebble relies on having else things go wonky). That's in addition to wanting to ensure users have agreed to Rebble TOS before using the service (after like... legal threats etc).

hellcp:
We can provide you with assurance that we will not pursue litigation for not requiring user login in core mobile app if you would feel more comfortable merging this. We would also be happy to modify APIs as needed to provide information that you need in appstore

radostin04:
I believe that a refusal to implement this change constitutes a clear contradiction to Core Devices' stated interest. It was made apparant to the users that the reason for implementing the app store feed system and for Core Devices starting their own app store is to provide users and app developers choice in where to obtain and host apps, and to prevent Pebble apps from being controlled by one organization.

If this is the case, why does Coreapp unfavourably favour Core Devices' app feed, despite there being no technical or legal reason for that to be the case? If there will be multiple app store feeds, they should both be treated as first class citizens of the Pebble ecosystem.

Henry-Hiles:
> I believe that a refusal to implement this change constitutes a clear contradiction to Core Devices' stated interest. It was made apparant to the users that the reason for implementing the app store feed system and for Core Devices starting their own app store is to provide users and app developers choice in where to obtain and host apps, and to prevent Pebble apps from being controlled by one organization.
> 
> If this is the case, why does Coreapp unfavourably favour Core Devices' app feed, despite there being no technical or legal reason for that to be the case? If there will be multiple app store feeds, they should both be treated as first class citizens of the Pebble ecosystem.

I completely agree, a good summary. It would be great to get the Rebble app feed as a first class citizen of the core app.

PR #85, opened 2026-01-03
notifications: Add support for watch timed mute options (1 hour, Today)

asyba:
Implement the temporary mute functionality for notifications.
This allows to support the new mute options (1 hour, Today) that pebble has.

For the moment is just for sync the status between databases.

PebbleOS PR: https://github.com/coredevices/PebbleOS/pull/651


libpebble3/schema/io.rebble.libpebblecommon.database.Database/27.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
```
sjp4, line 2: 27.json shouldn't be changed in this PR - that's why you were seeing migration problems (you probably did a build which changed the schema before changing the database version to 28). You'll need to rebase on the latest main (db version is now 33), and revert these changes to the schema json, then update the version to 34, compile, and checking those json changes

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/NotificationHandler.kt
```diff
@@ -149,11 +149,23 @@ class NotificationHandler(
         }
         val anyContactMuted = notification.people.any { it.muteState == MuteState.Always }
         val anyContactStarred = notification.people.any { it.muteState == MuteState.Exempt }
+        // Check if mute has expired
```
sjp4: This code path is never hit, right? (because there's no way to configure the temperary mute on android)

libpebble3/schema/io.rebble.libpebblecommon.database.Database/27.json
asyba: I see now how I have to deal with database changes, thanks for explain to me.

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/NotificationHandler.kt
asyba: No, we can't set the temporary mute on the app for the moment.

Only from the watch, but for Android the filter to decide if the notification should be sent or not it's done in the android app, so that is why I added this in this class.
but I don't have an android to test this specifically, maybe it's not used and needs to be removed??

Compare to ios that the filter decide can be done only in the watch since notification are sent directly to there.

sjp4: Yes - we never sync that database with the watch on android, so these could never be set (unless there was a UI built to do that)

asyba: done removed.

PR #87, opened 2026-01-04
screenshot: Add share and trigger buttons and support color

asyba:
* Enable iOS screenshot. same code from https://github.com/coredevices/mobileapp/pull/80
* Remove automatic trigger of taking a screenshot in that case there is a new button "Screenshot" and the same one is used to refresh.
* New Share button to share the image with the default OS share system.
* Support for color watch.

Only tested on iOS.

image

image

image

image


sjp4 (approved):
This is great, thanks!

PR #95, opened 2026-01-08
notifications: add enable sort and asc/desc order

asyba:
* Added a new sorter called "Enabled" to order notification list by Toggle enable.
* Added Ascending and Descending order, for example if you click order by Name it will order like a,b,c etc but if you click again order will be by z,y,x etc , the same for Recent, Enabled.  (not sure for "Count" since that is not supported by iOS) 

Tested iOS only


asyba:
cc @sjp4

sjp4 (approved):

PR #99, opened 2026-01-12
notifications: iOS support for custom icon/vibe/color

asyba:
iOS support for custom icon/vibe/color notifications.

The custom vibe pattern is not supported for the moment, it requires more work, not sure what is the best approach.
It requires an update to the PebbleOS firmware if not this will not take effect on the watch for iOS.
PR: https://github.com/coredevices/PebbleOS/pull/692

Tested on iOS only.

image
image

### DIFF
--- a/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
+++ b/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
@@ -22,7 +22,9 @@ import io.rebble.libpebblecommon.structmapper.StructMapper
 import io.rebble.libpebblecommon.util.DataBuffer
 import io.rebble.libpebblecommon.util.Endian
 import kotlinx.serialization.Serializable
+import io.rebble.libpebblecommon.timeline.toPebbleColor
 import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.runBlocking
 import kotlin.time.Instant
 
 @Immutable
@@ -55,14 +57,36 @@ data class NotificationAppItem(
         SFixedString(StructMapper(), packageName.length, packageName).toBytes()
 
     override fun value(params: ValueParams): UByteArray? {
-        val m = StructMapper()
-        val entity = NotificationAppBlobItem(
-            attributes = attributes {
-                appName { name }
-                muteDayOfWeek { muteState.value }
-                lastUpdated { stateUpdated.instant }
-            }.map { it.asAttribute() }
-        )
+        val attributesList = attributes {
+            appName { name }
+            muteDayOfWeek { muteState.value }
+            lastUpdated { stateUpdated.instant }
+            
+            vibePatternName?.let { name ->
+                val pattern = params.vibePatternDao?.let { dao ->
+                    runBlocking {
+                        dao.getVibePattern(name)
+                    }
+                }
+                pattern?.let {
+                    vibrationPattern { it.pattern }
+                }
+            }
+            
+            colorName?.let { name ->
+                io.rebble.libpebblecommon.timeline.TimelineColor.findByName(name)?.let { color ->
+                    backgroundColor { color.toPebbleColor() }
+                }
+            }
+            
+            iconCode?.let { code ->
+                io.rebble.libpebblecommon.packets.blobdb.TimelineIcon.fromCode(code)?.let { icon ->
+                    icon { icon }
+                }
+            }
+        }.map { it.asAttribute() }
+        
+        val entity = NotificationAppBlobItem(attributes = attributesList)
         return entity.toBytes()
     }
 
--- a/pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationsScreen.kt
+++ b/pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationsScreen.kt
@@ -130,9 +130,8 @@ fun NotificationAppCard(
     clickable: Boolean,
     showBadge: Boolean,
 ) {
-    val reallyClickable = clickable && platform == Platform.Android
     val app = entry.app
-    val modifier = if (reallyClickable) {
+    val modifier = if (clickable) {
         Modifier.clickable {
             nav.navigateTo(PebbleNavBarRoutes.NotificationAppRoute(app.packageName))
         }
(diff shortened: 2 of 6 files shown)


libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
```diff
@@ -57,26 +60,74 @@ data class NotificationAppItem(
...
+            buildVibePatternAttribute(m),
+            buildColorAttribute(m),
+            buildIconAttribute(m)
         )
+
+        val entity = NotificationAppBlobItem(attributes = attributes)
         return entity.toBytes()
     }
 
+    private fun buildVibePatternAttribute(m: StructMapper): Attribute? {
```
sjp4: I noticed that we weren't using the same attribute builder here that we are elsewhere in the app (hence lots of code here to reproduce what is already done elsewhere). I've updated the code to use that - if you rebase, hopefully you will be able to just re-use existing build methods (e.g. just calling `vibrationPattern {}` like [here](https://github.com/coredevices/mobileapp/blob/1d8023bc0a73302b28b64df9e1034cd98970826d/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotification.kt#L106)). Ideally everything should go through those APIs when creating attributes (sorry this didn't previously)

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
```diff
@@ -164,16 +215,37 @@ fun DbWrite.asNotificationAppItem(): NotificationAppItem? {
 //            logger.e("lastUpdated is null")
 //            return null
 //        }
+        
+        val iconCode = item.attributes.get(TimelineAttribute.Icon)?.let { bytes ->
```
sjp4: The watch can't modify any of these new fields, right? In which case I wouldn't write the code to deserialize them - just always set them all to `null` here and always use the existing values in `handleWrite()` in the DAO

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
```diff
@@ -57,26 +60,74 @@ data class NotificationAppItem(
...
         )
+
+        val entity = NotificationAppBlobItem(attributes = attributes)
         return entity.toBytes()
     }
 
+    private fun buildVibePatternAttribute(m: StructMapper): Attribute? {
+        return vibePatternName?.let { name ->
+            // Try to find default pattern first.
+            // TODO: Support custom patterns (requires DB access or passing full pattern here)
```
sjp4: This should use the `VibePatternDao` to read all patterns (i.e. not use `DefaultVibePattern` directly at all). I realize that's tricky right now - I just made a change to make this easier; you'll be able to add the dao to `ValueParams`, and populate that after injecting it in `BlobDb`

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationsScreen.kt
```diff
@@ -130,7 +130,7 @@ fun NotificationAppCard(
     clickable: Boolean,
     showBadge: Boolean,
 ) {
-    val reallyClickable = clickable && platform == Platform.Android
+    val reallyClickable = clickable
```
sjp4: probably can get rid of the `reallyClickable` variable

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
```diff
@@ -57,26 +60,74 @@ data class NotificationAppItem(
...
+            Attribute(TimelineAttribute.MuteDayOfWeek.id, SUByte(m, muteState.value).toBytes()),
+            Attribute(
+                TimelineAttribute.LastUpdated.id,
+                SUInt(
+                    m,
+                    stateUpdated.instant.epochSeconds.toUInt(),
+                    endianness = Endian.Little
+                ).toBytes()
+            ),
+        ) + listOfNotNull(
```
sjp4: Did you test this code with an old firmware (e.g. on a pebble time watch or something not-recent)? Asking because the initial android change wound up crashing the watch adding the vibe pattern attribute. We had to add a capability there which gated the phone adding the attribute. If it works, this is fine (it's a different code path on the watch etc so maybe it doesn't hit the same crash), but definitely needs testing.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
asyba: yes  we cannot edit them on the watch, I was thinking that doing something in the _handleWrite()_ was not the best option so I was expecting that at some point we should deserialize, but if we can do that, yes no point of doing this.
changed to return null

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
asyba: thanks I think I did it.

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationsScreen.kt
asyba: ok removed

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
asyba: No I did't test it, but after this changes I tested on my pebble time with old FW and iphone xr and it did't crash.
the watch gets default icon/color/vibration not matter what the phone app sets its ignore them.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
asyba: Thanks done.

sjp4 (approved):

Issue #46, opened 2025-11-27
Future of libpebble3


matejdro:
With the release of the full app source code, I'm wondering what does that mean for the future of the libpebble3? Will the library still be updated?

If that is the case, then ideally, it would be in its own repo, so everyone including it does not need to checkout the entire app.

sjp4:
Do you want to add the extra settings.gradle file @matejdro ?

PR #112, opened 2026-02-08
Add Timeline system app

matejdro:
PR adds timeline app (https://github.com/coredevices/PebbleOS/pull/708) to the system apps, so it can be sorted.

### DIFF
--- a/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/locker/SystemApps.kt
+++ b/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/locker/SystemApps.kt
@@ -20,7 +20,7 @@ enum class SystemApps(
     Health(SystemAppIDs.HEALTH_APP_UUID, "Health", AppType.Watchapp, WatchType.entries, defaultOrder = -5),
     Workout(SystemAppIDs.WORKOUT_APP_UUID, "Workout", AppType.Watchapp, WatchType.entries, defaultOrder = -4),
     Watchfaces(SystemAppIDs.WATCHFACES_APP_UUID, "Watchfaces", AppType.Watchapp, WatchType.entries, defaultOrder = -3),
-    Timeline(SystemAppIDs.TIMELINE_MENU_ENTRY_UUID, "Timeline", AppType.Watchapp, WatchType.entries, defaultOrder = -3),
+    Timeline(SystemAppIDs.TIMELINE_MENU_ENTRY_UUID, "Timeline", AppType.Watchapp, WatchType.entries, defaultOrder = 100),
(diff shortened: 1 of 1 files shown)


libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/locker/SystemApps.kt
```diff
@@ -20,6 +20,7 @@ enum class SystemApps(
     Health(SystemAppIDs.HEALTH_APP_UUID, "Health", AppType.Watchapp, WatchType.entries, defaultOrder = -5),
     Workout(SystemAppIDs.WORKOUT_APP_UUID, "Workout", AppType.Watchapp, WatchType.entries, defaultOrder = -4),
     Watchfaces(SystemAppIDs.WATCHFACES_APP_UUID, "Watchfaces", AppType.Watchapp, WatchType.entries, defaultOrder = -3),
+    Timeline(SystemAppIDs.TIMELINE_MENU_ENTRY_UUID, "Timeline", AppType.Watchapp, WatchType.entries, defaultOrder = -3),
```
sjp4: Behaviour here will be: timeline gets inserted at the top of the apps list after users upgrade to this version.

For previous changes like this I've forced it to reset the position of all system apps each time (by changing PREF_KEY_HAVE_INSERTED_SYSTEM_APPS_AT_CORRECT_POSITION). I don't love that - we probably need something better - what do you think?

matejdro: Hm. Would it be better to insert them at the bottom?

matejdro:
Hm, it also appears we would need to add an image for this to the `pebble/src/commonMain/composeResources/drawable`. I am not really a designer, could you provide an image for this?

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/locker/SystemApps.kt
sjp4: I'm honestly not sure - that might be easiest (just set it as e.g. position 100)

matejdro:
Also, this is another reason, why I would put libpebble3 into a separate repo (#46). Now, everyone contributing to it, also has to fix all the incompatibilities generated in the app

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/locker/SystemApps.kt
matejdro: Changed

PR #119, opened 2026-02-15
iOS Shortcuts (notifications, quiet time, watch prefs, launch app)

asyba:
iOS support for Shortcuts.
Some possible actions that could be used, for example using Focus modes.

I also added some classes but not implement, to support the equivalent in Android in the future using the Tasker plugin or others.

IMG_8097
IMG_8097
IMG_8098


composeApp/src/commonMain/kotlin/coredevices/coreapp/di/utilModule.kt
```diff
@@ -70,4 +71,5 @@ val utilModule = module {
     single { CoreConfigHolder(defaultValue = CoreConfig(), get(), get()) }
     single { CoreConfigFlow(get<CoreConfigHolder>().config) }
     singleOf(::ModelManager)
+    singleOf(::CactusTranscriptionService)
```
asyba: I have to add this because it was crashing the app when installing, maybe its a bug not yet discovered.

ericmigi:
really cool! Really useful

saltedlolly:
Awesome. I was hoping for Shortcuts support at some point! Thanks for adding it!

Would it be possible to add a timeline pin from a shortcut?

composeApp/src/commonMain/kotlin/coredevices/coreapp/di/utilModule.kt
sjp4: cc @crc-32 - I guess this is in experimentalmodule, but references from common code?

asyba:
Added timeline shortcuts
thanks @saltedlolly good idea.

IMG_8111
IMG_8112

iosApp/iosApp/PebbleShortcuts.swift
```diff
@@ -0,0 +1,771 @@
...
+        } else {
+            data = Data()
+        }
+        let file = IntentFile(data: data, filename: "watch_screenshot.png")
+        return .result(value: file)
+    }
+}
+
+@available(iOS 16.0, *)
+struct PebbleShortcutsProvider: AppShortcutsProvider {
```
sjp4, line 691: Does this code need to be in swift, or was it just easier to to write in swift? Our policy is basically that it should be in kotlin unless the only way to do it is in swift (or to put it another way, as much in kotlin as is possible). I don't have much context on how shortcuts work (or the ability to thoroughly review swift code), so any context would be useful

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt
```diff
@@ -423,5 +424,10 @@ class LibPebble3(
             val libPebble = koin.get<LibPebble>()
             return libPebble
         }
+
+        /**
+         * Returns the [RemoteTimelineEmulator] from the LibPebble Koin context.
+         */
+        fun getRemoteTimelineEmulator(): RemoteTimelineEmulator = koin.get()
```
sjp4: `RemoteTimelineemulator` shouldn't be exposed like this (it's internal to libpebble) - we should add new methods to the libpebble API (probably [here](https://github.com/coredevices/mobileapp/blob/b26287b4e44f4e0632a6e6dda71e6b31c3bb761a/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt#L133)) if the ones already exposed don't do the job

pebble/src/androidMain/kotlin/coredevices/pebble/actions/PebbleActions.android.kt
```diff
@@ -0,0 +1,112 @@
...
+
+import io.rebble.libpebblecommon.connection.LibPebble
+
+/**
+ * Android implementation of Pebble actions.
+ *
+ * For now these are placeholders (no-op) but provide a clear extension point
+ * for future integrations (e.g. Tasker plugin, Android Shortcuts, etc.).
+ */
+class AndroidPebbleAppActions(
```
sjp4: If we add anything similar on android it will be part of PebbleKit 2.0, so I think it's unlikely we'd re-use this interface directly. That being said, i'm thinking it's probably best to keep it entirely within the `iosMain` code tree for now (i.e. no `androidMain` implementations needed)

sjp4 (commented):
General questions:
- Should these be documented as an "API", or is that not really shortcuts work?
- Should they always be enabled, or would users want to enable/disable access to this stuff?

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt
asyba: the link you posted is the same place I added getRemoteTimelineEmulator
libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt

pebble/src/androidMain/kotlin/coredevices/pebble/actions/PebbleActions.android.kt
asyba: okay. I removed the `androidMain` implementations, there are still `commonMain` files let me know if those should be move to `iosMain` tree.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/LibPebble.kt
asyba: done, found another way. thanks

iosApp/iosApp/PebbleShortcuts.swift
asyba: No, is not possible to do it with kotlin, its specific swift. 
I improve it by moving some methods/code to kotlin but this is as much that can be done.

in this section func `perform() ` is wheere each shorcut action calls kotlin code (`IOSDelegateShortcuts.shared`).

pebble/src/androidMain/kotlin/coredevices/pebble/actions/PebbleActions.android.kt
sjp4: I think we should move those to `iosMain`

asyba: done

sjp4 (approved):

alexrmay91:
> I also added some classes but not implement, to support the equivalent in Android in the future using the Tasker plugin or others.

I didn't see this in the discussion but Android has native app shortcuts too. Unless they're lacking in functionality, this path seems like the best one as it's similar to the iOS implementation and tasker can access these already but so many other apps can too (launcher gestures, pinning shortcuts to your home screen, tap and hold on the app, probably a lot more).

I love tasker, but I know a lot of people won't use it. Native shortcuts make it accessible to the average user.

[https://developer.android.com/develop/ui/views/launch/shortcuts/creating-shortcuts](https://developer.android.com/develop/ui/views/launch/shortcuts/creating-shortcuts)

PR #121, opened 2026-02-19
Add Japanese (ja_JP) language packs

jamsinclair:
## Summary

I know you're not actively receiving contributions, but it would be awesome to add ja_JP support 🙏 

Adds two Japanese language packs sourced from: https://github.com/elliottback/PebbleTimeJapaneseLanguagePack
- Regular font weight language pack
- Light font weight language pack

Attributions given to the author Kuro.


pebble/src/commonMain/kotlin/coredevices/pebble/services/LanguagePackRepository.kt
```diff
@@ -1845,6 +1845,34 @@ private val LanguagePacksJson = """
       "localName": "català",
       "name": "Catalan - by MarSoft",
       "version": 1
+    },
+    {
+      "ISOLocal": "ja_JP",
+      "file": "https://github.com/elliottback/PebbleTimeJapaneseLanguagePack/raw/1e914c39dc459c03ce8a1ae6ee7c17f59f56f21c/pblp_zhs_zht_ja_v5_regular.pbl",
+      "firmware": "4.0.0",
+      "hardware": null,
```
jamsinclair, line 1853: I think given the file size of the language pack early Pebble devices with limited memory may not support this. Any guidance on how we can configure the appropriate devices via the metadata? The hardware names seem to not match the official platform names [seen in the docs](https://developer.repebble.com/guides/tools-and-resources/hardware-information/). Is there wiki or resource with these hardware names defined?

sjp4: The codes are the ones used [here](https://github.com/coredevices/mobileapp/blob/aa17e91650a4fb7ba8ba67205c91637563a71d46/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/metadata/WatchHardwarePlatform.kt#L12) - I think you'd need to duplicate the json for each supported platform if you don't want to put `null` (all platforms)

jamsinclair: Amazing, thank you @sjp4.

Thinking off the top of my head, would there be any aversion to making "hardware" to optionally accept an array? E.g. Spanish (es_ES) is duplicated 6 times. This could be defined once with:
```json
{
  "ISOLocal": "es_ES",
  "file": "https://binaries.rebble.io/lp/u7f2giK-es_ES.pbl",
  "firmware": "3.8.0",
  "hardware": ["ev2_4", "snowy_s3", "v2_0", "snowy_dvt", "v1_5", "spalding"],
  "id": "5670a5c2b9abe4160010c224",
  "localName": "Español",
  "mobile": {
    "name": "ios",
    "version": "2.6.0"
  },
  "name": "Spanish",
  "version": 28
}
```

Happy to contribute this change, likely it's just a small serializer change and updating how language packs are filtered for display.

sjp4: Sorry for the delay - the only thing with changing the field is that this is modelled on the old pebble LP endpoint response (now also used by rebble), so it would be nice to maintain compatibiltiy with that (i.e. by not changing a string to a list<string>. if you can figure out a way around that (e.g. by adding a new field to do this, instead of changing the existing one), then feel free!

jamsinclair: Sorry for my delay 🙇 

Circling back to this I think let's just leave it as null as a starting point. It is likely going to lesser used locale and if causes enough issues we can add the relevant duplicates with hardware.

jamsinclair:
Apologies for the delay @sjp4. Let's avoid scope creep and keep this addition simple to start with. Ready for review again 🙏

sjp4 (approved):

PR #124, opened 2026-02-21
calls: fix call duration timer started for STATE_DISCONNECTING

spr4bhu:
This PR fixes [coredevices/PebbleOS#808](https://github.com/coredevices/PebbleOS/issues/808) — abnormal call durations and incorrect status text for unanswered calls on Pebble smartwatches.

When an incoming call is not answered on a Pebble watch, the watch sometimes displays an abnormal call duration (e.g., `491791:41:09`). The root cause is in LibPebbleInCallService.kt, which maps both Call.STATE_ACTIVE and Call.STATE_DISCONNECTING to ActiveCall. When a call disconnects before being answered, the old timestamp persists and gets used to compute a garbage duration across the entire idle period.

Changes:
Remove Call.STATE_DISCONNECTING from the ActiveCall mapping — only truly active calls now trigger the duration timer. Calls that disconnect before being answered skip the timer entirely.

### DIFF
--- a/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/calls/LibPebbleInCallService.kt
+++ b/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/calls/LibPebbleInCallService.kt
@@ -121,7 +121,7 @@ class LibPebbleInCallService : InCallService(), LibPebbleKoinComponent {
                 onCallEnd = { call.disconnect() },
                 cookie = cookie,
             )
-            Call.STATE_ACTIVE, Call.STATE_DISCONNECTING -> io.rebble.libpebblecommon.calls.Call.ActiveCall(
+            Call.STATE_ACTIVE -> io.rebble.libpebblecommon.calls.Call.ActiveCall(
                 contactName = call.resolveContactName(),
                 contactNumber = call.resolveContactNumber(),
                 onCallEnd = { call.disconnect() },


libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/calls/LibPebbleInCallService.kt
```diff
@@ -121,7 +121,7 @@ class LibPebbleInCallService : InCallService(), LibPebbleKoinComponent {
                 onCallEnd = { call.disconnect() },
                 cookie = cookie,
             )
-            Call.STATE_ACTIVE, Call.STATE_DISCONNECTING -> io.rebble.libpebblecommon.calls.Call.ActiveCall(
```
sjp4, line 124: I didn't fully understand how this triggers the bug - seems like a sensible change regardless, but would be nice to understand why it happened (should it also remove the active call when it is `STATE_DISCONNECTING`?)

spr4bhu: STATE_DISCONNECTING was previously mapped to ActiveCall, which sent PhoneControl.Start to the watch - starting the duration timer for a call that was never actually connected. The stale call_start_time then produced garbage durations like 491791:41:09.

Returning null for STATE_DISCONNECTING effectively removes the active call. PhoneControlManager observes currentCall and when it becomes null, it uses the saved lastCookie to send PhoneControl.End to the watch, then clears it. This means STATE_DISCONNECTING can now end calls on the watch, without using onCallRemoved. iOS unanswered calls are handled separately on the firmware side.

PR #135, opened 2026-03-16
Add support for notification filtering by regex

aveao:
This PR adds this cute little feature so you can filter notifications by regex. It applies to both body and title, and is currently case sensitive (tho `(?i)` works). lmk if you want either changed, best would be if it could have switches but tbh that seems like so much UI and designing good UI isn't my expertise.

It also makes the entire notification section scrollable instead of just the notification categories, which makes it so that you can scroll when you've added a few regexes (it's also a nicer way to interact overall imo).

The regexes are saved on ON_STOP (switching apps, locking phone, home button), onDispose (navigating out of the view) and focus loss (tapping another field or closing keyboard).

lmk if you want anything squashed in any particular way.

image

(Only tested on android)


aveao:
Hi Steve, thanks for your reply.

> I've been thinking that the framework for how it works behind the scenes should be more flexible than "regexes per-app" (i.e. should handle a global filter which doesn't care which app).

I do quite a bit of notification filtering on my phone in general, but never felt the need to have a global filter. What do you have in mind?

> I think that for "normal" users we need something simpler than regex as the user interface (i.e. "includes"/"excludes") - maybe with the option to also use regex as a more "advanced" feature.

For the sake of UI and simplicity, I could move it to just a text include/exclude then add a setting to near the bottom of the app settings to treat them as regexes.

> For the UX, I've been thinking that the best way to interact with the feature would be via the notification history brower (i.e. select a past notification that you'd want to hide in the future, use UI to drag-to-select the bit that you'd want to match on, and click to create a "rule" from that. That's kinda advanced, so we could totally build the feature and add that later.

This seems a bit clunkier for me but I'm also not the average user you're aiming for :) Can definitely see it being easier for some. I'd say that it may be best to leave it to a separate PR.

> Given that, I was envisaging a more generic model - something like a new DB table "notification_rules" where each row has a "target_type" (app/global/maybe contact), "target" (app/contact id if required for the type), rule_type (e.g. include_text/exclude_text/include_regex/etc). Could also be e.g. body_include_text - i'm not really sure what the granularity should be there). You can see in NotificationScreen there are some commented-out references to a history screen and a rules screen, but nothing else ever happened there.

(see note about global above)
Are contact IDs stable across apps?
I do think I want to add granularity of title vs text.

Maybe I'll swap the add/edit ui to a new modal that just has checkboxes and dropdowns to set the regex, what fields it applies to, etc. That would make it easier from the UI perspective.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
```diff
@@ -205,6 +207,7 @@ fun DbWrite.asNotificationAppItem(): NotificationAppItem? {
             vibePatternName = null,
             colorName = null,
             iconCode = null,
+            filterIsAllowlist = false,
```
sjp4: I do think we should move this to the rules table - it's more flexible, and at the same time less complicated (not having to load rules from multiple places to resolve them).

> I ended up leaving the allowlist option within the app level as having include/exclude conflicts is rather annoying, doing an update over all items or just blocking off the option to add the other rule type if there's already a rule didn't really result in a nice experience.

I think we have to handle conflicts (i.e. if there are allow and deny rules that both apply to a message, then it has to satisfy all of them - match the allow as well as not-match the deny).

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationRuleEntity.kt
```diff
@@ -0,0 +1,43 @@
...
+    @PrimaryKey(autoGenerate = true) val id: Long = 0,
+    val targetType: TargetType,
+    val target: String?,
+    val matchType: MatchType,
+    val matchField: MatchField,
+    val pattern: String,
+    val caseSensitive: Boolean = false,
+)
+
+enum class TargetType(val value: Int) {
```
sjp4: I'd prefer string codes for all the nums (e.g. `val value: String`/`App("app")` so that the db table is readable in app inspection

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
```diff
@@ -105,125 +124,339 @@ fun NotificationAppScreen(
...
         }
     }
 }
 
+@Composable
+private fun NotificationRulesSection(
+    app: NotificationAppItem,
+    notificationApps: NotificationApps,
+) {
+    val rules by notificationApps.notificationRulesForApp(app.packageName).collectAsState(emptyList())
```
sjp4: I think that probably needs a `remember` on the result of the query just to avoid re-running the db query every composition e.g. 

```
val query = remember(app) { notificationApps.notificationRulesForApp(app.packageName).}
val rules by query.collectAsState(emptyList())
```

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/dao/NotificationRuleDao.kt
```diff
@@ -0,0 +1,27 @@
...
+@Dao
+interface NotificationRuleDao {
+    @Query("SELECT * FROM NotificationRuleEntity WHERE targetType = :targetType AND target = :packageName")
+    fun getRulesForApp(targetType: TargetType = TargetType.App, packageName: String): Flow<List<NotificationRuleEntity>>
+
+    @Query("SELECT * FROM NotificationRuleEntity WHERE targetType = :targetType AND target = :packageName")
+    suspend fun getRulesForAppOnce(targetType: TargetType = TargetType.App, packageName: String): List<NotificationRuleEntity>
+
+    @Insert
+    suspend fun insert(rule: NotificationRuleEntity): Long
```
sjp4: nit: could combine insert/update with an `@Upsert` method?

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
```diff
@@ -105,125 +124,339 @@ fun NotificationAppScreen(
...
+    existing: NotificationRuleEntity?,
+    onDismiss: () -> Unit,
+    onSave: (NotificationRuleEntity) -> Unit,
+) {
+    var matchType by remember { mutableStateOf(existing?.matchType ?: MatchType.Text) }
+    var matchField by remember { mutableStateOf(existing?.matchField ?: MatchField.Both) }
+    var pattern by remember { mutableStateOf(existing?.pattern ?: "") }
+    var caseSensitive by remember { mutableStateOf(existing?.caseSensitive ?: false) }
+
+    val regexError = remember(pattern, matchType) {
```
sjp4, line 407: Nice!

sjp4 (commented):
Overall, looks great! Left a few comments

Other ideas (feel free to implement/discard):
- Probably would be nice to enable/disable rules, rather than simply deleting them
- We might want to consider making the rules section in the UI collapsible if user has added a lot of rules (probably not many people will do this..)

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
dequis: Hello! I'm ave's girlfriend, I requested her to implement this feature for me because I missed it from gadgetbridge, and I have opinions.

* I think UI/UX is about tradeoffs and having maximum flexibility is not necessarily good. The modal in the latest commits gives us a lot more room to add features, but I'd argue we've used up all that room already.
* There's a cost in this flexibility. The software may be able to handle conflicts, but the user needs to handle it too
* Use cases are important: I think I could sort of come up with a use case that makes this feature worth implementing, but I don't have anything concrete tbh.
* "match the allow as well as not-match the deny" is, honestly, confusing. This can easily create unintended undefined states. I don't think it's easy to communicate in the UI. It's also not clear what the action is when no rules apply to a notification - that sounds to me like two separate undefined states. You really want to avoid exposing users to that situation.
* My preferred mental model for processing rules with allow/deny with each other is a firewall in the style of iptables:
    * rules are processed sequentially
    * each rule has an action, accept or drop
    * if the rule matches the action is final
    * if the rule doesn't match, it continues to the next rule
    * you go through the rules and apply their actions
* I personally think this mental model is easy enough to understand, given two features:
    * The "policy", the application-wide fallback decision when no rules apply (the filterIsAllowlist we already have)
    * A clear feature to reorder rules (this one's heavy on the UI side)

In general, you really want to have something like filterIsAllowlist at the app level, there's no getting away from the fact that filters won't catch everything.

For now I would prefer if we keep it in the current state. I think the use cases are niche and this should be more than enough to make power users happy. Gadgetbridge doesn't even have regexes, and gadgetbridge is free software licensed under the AGPLv3 and distributed on f-droid.

Also, if you wish to add it later, the current database structure allows enough flexibility to have allow/deny in individual rules and to make sorting explicit. It does require a proper think in UI/UX to pull off, I'm glossing over details in this feature. Which is why I'd rather keep it simple.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/dao/NotificationRuleDao.kt
aveao: Addressed here: https://github.com/coredevices/mobileapp/pull/135/commits/a3237b49a59ed7d238237577704120fad8d5677c

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
aveao: Addressed here: https://github.com/coredevices/mobileapp/pull/135/commits/4ef2fed582dbc4019e20a49d7d8ffd2c306e40e2

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationRuleEntity.kt
aveao: Oh that's pretty neat, I did end up using enums but with strings instead. This should be reasonable within DB I hope, it's how many other fields appear to be stored.

https://github.com/coredevices/mobileapp/pull/135/commits/2f8274554001d40dd02eff8f08fc47aaf01cb314

aveao:
Addressed the feedback except for the one pointed out by dx. Did some other fixes (see commit text) and touched up the UI to be a bit nicer.

---

Better spacing, add button more in-line with current design, switched to row/column to have buttons properly centered and to have a nicer visual style:

<details>

<summary>big screenshot</summary>

Screenshot_20260401_224907

</details>

Hid block/allow if there's no rules, and inserted a "No notification rules" text: (no rules are attempted to be applied if there's none added for that app, so if app is set to "allow if matching" when there's no rules, it doesn't block all notifications for that app. so I think hiding is alright UI.)

<details>

<summary>big screenshot</summary>

Screenshot_20260401_223348

</details>

Also made it so that it is hidden on iOS (can be made visible later when fw adds support + fw is validated as a version supporting it or smth).

---

re your other ideas:

> Probably would be nice to enable/disable rules, rather than simply deleting them
> We might want to consider making the rules section in the UI collapsible if user has added a lot of rules (probably not many people will do this..)

Agreed on both, former IDK about how to best achieve UI. Both could be done in future PRs I think, this is already somewhat sizable :)

---

Should prob squash sooner than later, but going to hold off. I believe github gives the contributors to this repo access to the branch, so also feel free to squash as you see fit at any point.

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/NotificationAppEntity.kt
sjp4: > My preferred mental model for processing rules with allow/deny with each other is a firewall in the style of iptables:

I'm not at all convinced that iptables is a good UX/easy to understand mental model, as an example to follow.

> Also, if you wish to add it later, the current database structure allows enough flexibility to have allow/deny in individual rules and to make sorting explicit. It does require a proper think in UI/UX to pull off, I'm glossing over details in this feature. Which is why I'd rather keep it simple.

It doesn't though - if we have an allow/deny switch at the app level, and have rules that can apply at the app level or at a global level in the future, and we can't add an allow/deny property to rules in the future because that would become crazy complicated (allow/deny at multiple levels applying to the same rule?).

I see 2 option in the short term:
- Move allow/deny from the app to the rule, and figure out what that means for the UX. I'm actually not sure what the original use-case was, behind adding allow/deny at the app level, so i'm not sure how badly we actually need this.
- Remove allow/deny - everything is just configured in the rules table (with the flexibility of rules allowing includes vs excludes patterns), all as deny. Much simpler and not causing our future selves problems.

FWIW I think whatever is implemented (even basic deny filters) will be somewhat confusing until UX is developed a bit, but that's OK because most users won't touch this feature. What i'm thinking is that when adding a rule, the user can see (using the real notification history we have stored for that app) which messages would be filtered out by each rule. The history can also show exactly which rule filtered out a given notification. That's kind of the main reason I added the history screen a while back (that and allowing users to screenshot anomalies when they send support requests).

aveao: Here's my thoughts:
- Allow + deny coexisting is confusing UX, so I don't think it should be per-rule
- I don't see a good usecase for global. Do you have any ideas in mind?
- If there's both global and per-app rules, it could perform either as two-stage check (e.g. filter out all notifications mentioning work, then per-app you can do allow/deny on remaining notifs) or overriding global rules when rules on app level are set.
- I'd be happy to implement the feature to e.g. add a button to view what notifications this rule / all rules would match, but I would rather make that a separate PR to keep this one to a reasonable size.

sjp4: > I don't see a good usecase for global. Do you have any ideas in mind?

We already have this (contacts filtering). It's only not implemented as a generic rule because we didn't have those yet, but now it would be (and we'll probably still need to use "global" rules for this in the case where apps don't correctly link notifications to contacts, to manually do it by name). They even have both allow and deny.

I would also likely setup rules so that any mention of <thing x> from any app gets delivered to my watch (I don't want to miss out of getting tickets for my favourite band and I don't know whether the notification for them would come through email or instagram or..), or any mention of <thing y> doesn't get delivered.

> I'd be happy to implement the feature to e.g. add a button to view what notifications this rule / all rules would match, but I would rather make that a separate PR to keep this one to a reasonable size.

Definitely not in scope for this PR

aveao: I got rid of the allow functionality entirely as it didn't seem like it had many usecases, and complicated ux and logic too much.

sjp4 (approved):

aveao:
Neat! Thanks for the in-depth review. I'll open one or two more PRs in the next days, one for the theming of the categories, and one for a log viewer.

Screenshot_20260415-125052

Screenshot_20260415-125059~2

PR #146, opened 2026-03-31
Fix DictationResult serialization when sending UTF-8 text

ClusterM:
I speak languages that use the Cyrillic script, and whenever I used speech recognition, the output was always garbled: half of each word was cut off.

I finally found why: it’s the serialization between the app and the watch. We used string length in characters where we should use UTF-8 length in bytes — a classic mistake. English speakers wouldn’t really hit this bug; for non‑Latin languages it’s a serious issue.


sjp4 (approved):

DX37:
I've been waiting for this for a while. Will test when new version of the app is published. Thanks!

PR #149, opened 2026-04-06
Add Notifications Mute status

asyba:
* A bug was fixed where the mute toggle was not reflecting the correct status when the muteExpiration setting had an expired date and was subsequently reactivated.

* A status was added to the notification app view that indicates whether the mute is always active, active on weekends/weekdays, or only active on specific hours/days.

IMG_1999
IMG_1992
IMG_1994

### DIFF
--- a/pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
+++ b/pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
@@ -10,6 +10,8 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.Spacer
+import kotlin.time.Clock
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
 import androidx.compose.material.icons.Icons
@@ -189,7 +191,7 @@ fun NotificationAppScreen(
                             }
                         }
                     }
-                    LazyColumn {
+                    LazyColumn(modifier = Modifier.weight(1f)) {
                         items(
                             items = channelGroups,
                             key = { it.id },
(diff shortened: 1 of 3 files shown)


pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
```diff
@@ -218,6 +218,44 @@ fun NotificationAppScreen(
                             HorizontalDivider(modifier = Modifier.padding(top = 8.dp))
                         }
                     }
+                } else {
+                    androidx.compose.foundation.layout.Spacer(modifier = Modifier.weight(1f))
```
sjp4: should be able to import this

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
```diff
@@ -218,6 +218,44 @@ fun NotificationAppScreen(
...
+                }
+
+                val expiration = app.muteExpiration
+                val now = kotlin.time.Clock.System.now()
```
sjp4: import

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
```diff
@@ -218,6 +218,44 @@ fun NotificationAppScreen(
                             HorizontalDivider(modifier = Modifier.padding(top = 8.dp))
                         }
                     }
+                } else {
```
sjp4: Nesting here isn't very clear - might be better to have a separate if (iOS) { } block

asyba: done

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
asyba: done

pebble/src/commonMain/kotlin/coredevices/pebble/ui/NotificationAppScreen.kt
asyba: done

sjp4 (approved):

PR #154, opened 2026-04-15
Add a log viewer

aveao:
Looks like this:

Screenshot_20260415-125052

Screenshot_20260415-125059~2

Gated behind the debug options. I wanted this as sdk connection is broken for me, and the pkjs logs aren't accessible through any reasonable option (chrome dev connection then console works, but is a big hassle vs just checking a special section within the app).

### DIFF
--- a/composeApp/src/commonMain/kotlin/coredevices/coreapp/ui/navigation/AppNavHost.kt
+++ b/composeApp/src/commonMain/kotlin/coredevices/coreapp/ui/navigation/AppNavHost.kt
@@ -22,6 +22,7 @@ import coredevices.EnableExperimentalDevices
 import coredevices.ExperimentalDevices
 import coredevices.coreapp.ui.screens.BugReportScreen
 import coredevices.coreapp.ui.screens.BugReportsListScreen
+import coredevices.coreapp.ui.screens.LogViewerScreen
 import coredevices.coreapp.ui.screens.OnboardingScreen
 import coredevices.coreapp.ui.screens.ViewBugReportScreen
 import coredevices.coreapp.ui.screens.WatchOnboardingScreen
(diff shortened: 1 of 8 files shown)


ericmigi:
Not sure if we publish this anywhere, but you can actually download all of your own logs with this button. 
image

Probably will be best if we keep the log viewing out of the app, just for simplicity and not worrying about breaking it in future

aveao:
> Probably will be best if we keep the log viewing out of the app, just for simplicity and not worrying about breaking it in future

Gotcha. Happy to close this, but some thoughts:
- This utilizes `coredevices.coreapp.util.FileLogWriter` (also used by bug report system, incl. the export you mention), so as long as that remains with a somewhat stable API, this shouldn't break.
- Even if not in this shape, I think it'd be a very useful development tool down the line to be able to directly view the pkjs logs directly on device without the need for exporting logs, connecting via sdk (which silently fails for me), or using chrome dev connection to the pkjs thread. Debugging why an app worked on emulator but not on my phone was difficult to debug due to how long it took between change code -> compile -> install -> pull log cycles, which is what encouraged me to do this.
- I think the mimetype fix is a worthwhile addition and I'll send that as a separate PR on its own

aveao:
Oh and I'll probably keep maintaining this in my branch, as implied by #155 I'm interested in running a modified version on my phone with some added features that I don't think all would be interesting to upstream. I'm also a big fan of [custom personalized software](https://bsky.app/profile/did:plc:nro7kx7qz6fe4j3mezfqhd6m/post/3mjd4rly7wk2u). I'll still keep trying to upstream things as appropriate, do let me know if there's any better ways to coordinate what would be desired than just sending PRs, and also if it'd be welcome to contribute changes that make compiling the codebase function better (e.g. not defaulting to cloud transcription if relevant keys are not provided within build config).

ericmigi:
We're pretty responsive to issues here, so if you raise an issue, you'll get Steve, me, or Alice here to discuss it.

Issue #55, opened 2025-12-07
Build APKs in CI


julienrbrt:
It would be great to be able to download the APK of releases directly from the GitHub releases.
I am using [Obtainium](https://github.com/ImranR98/Obtainium) to manage my FOSS app and this one is missing.

PanderMusubi:
When APK files are released in this repo, it would also be possible to request inclusion in https://gitlab.com/IzzyOnDroid/repo , see also https://apt.izzysoft.de/fdroid/

eduncan911:
Found multiple issues on this missing signed APK in this repo.

Would the developers accept a PR to modify the CI to published a signed APK?

Us GrapheneOS users require it.

PR #155, opened 2026-04-15
Cohorts improvements

aveao:
Two parts to this PR:
- Always fall back to cohorts for updates if no memfault project key is provided. Relates to the comments within https://github.com/coredevices/mobileapp/issues/55
- Utilize cohorts API properly instead of using the source json in github.

Reasoning:

It's not great that the self-built versions of the app cannot receive updates for the coredevices watches, and the option with least code changes here to get there was to fall back to cohorts in that case. The updates are quite frequent over there from what I see (every few days) and so it'd be a big hassle to update manually, so I believe there's value in implementing this fallback.

I also took the liberty to work with rebble folks to make auth optional there (specifying is still useful for statistics), so you can e.g. [click here](https://cohorts.rebble.io/cohort?select=fw&hardware=silk&mobilePlatform=test&mobileVersion=test&mobileHardware=test&pebbleAppVersion=test) and get update info without authenticating now.

Now I'm working with them to make cohorts also serve the coredevices firmwares, and that's not likely to end up in the config.json with the way we've been brainstorming so far, so that required the second half of these changes.

### DIFF
--- a/pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
+++ b/pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
@@ -1,6 +1,8 @@
 package coredevices.pebble.firmware
 
+import CoreAppVersion
 import co.touchlab.kermit.Logger
+import coredevices.pebble.account.bootConfigPlatform
 import coredevices.pebble.services.HttpClientAuthType
 import coredevices.pebble.services.PebbleHttpClient
 import coredevices.pebble.services.PebbleHttpClient.Companion.get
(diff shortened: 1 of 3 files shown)


ericmigi:
super helpful, thanks @aveao!

pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
```diff
@@ -1,6 +1,8 @@
 package coredevices.pebble.firmware
 
+import CoreAppVersion
 import co.touchlab.kermit.Logger
+import coredevices.pebble.account.bootConfigPlatform
 import coredevices.pebble.services.HttpClientAuthType
```
sjp4, line 6: Maybe some imports to remove from the previous json code?

pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
```diff
@@ -14,36 +16,41 @@ import kotlin.time.Instant
...
-        val normalFw = response.hardware[watch.platform.revision]?.normal
+        val normalFw = response.fw?.normal
         if (normalFw == null) {
-            logger.i { "No firmware found for ${watch.platform.revision}" }
+            logger.i { "No firmware found for $hardware" }
             return FirmwareUpdateCheckResult.UpdateCheckFailed("Failed to check for PebbleOS update")
         }
         val latestFwVersion = FirmwareVersion.from(
-            tag = normalFw.version,
+            tag = normalFw.friendlyVersion.removePrefix("v"),
```
sjp4: I don't think you need to strip the "v" there - the parser handles that (previous implementation which used cohorts until a few weeks ago didn't do that)

sjp4 (commented):
Is this tested with e.g. a classic/time watch that updates still work?

pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
aveao: Ah great, will change

pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
aveao: Will do

aveao:
I did not test if the actual update succeeds, however it does load the metadata correctly. Can test an update early next week if you want a more in depth test.

Screenshot_20260416-085323

pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
aveao: done: https://github.com/coredevices/mobileapp/pull/155/commits/0a6c6ead6d64dfc9a1c908f8f91b0d2e12b82383

pebble/src/commonMain/kotlin/coredevices/pebble/firmware/Cohorts.kt
aveao: done: https://github.com/coredevices/mobileapp/pull/155/commits/0a6c6ead6d64dfc9a1c908f8f91b0d2e12b82383

sjp4 (approved):

PR #156, opened 2026-04-18
iOS: make single click and hold configurable (Index Agent or Webhook)

iFixRobots:
## Summary

Adds an iOS-only "Single click and hold" primary-mode picker. Experimental users can now choose between today's Index Agent behavior (default, unchanged) and a Webhook-only mode that skips the on-device agent. A "Share with Index agent" toggle re-enables the agent inside Webhook mode for users who want both.

## Motivation

Today the ring's primary gesture (single click and hold) is hardcoded to route every recording through the Index agent. Every transcription appears in the Index feed and is processed by the on-device LLM, which frequently generates low-value notes. Users have no way to opt an individual recording — or all primary-gesture recordings — out of this flow.

The double click and hold offers more options, but I'd like both single click and double click to be configurable. This PR extends the same underlying `IndexWebhookUploadRecordingOperation` to the primary gesture, behind the same `experimentalDevicesEnabled` gate, and adds a user-toggleable "Share with Index agent" flag that controls whether the agent runs on top of the webhook upload.

## Scope

**iOS only.** All UI additions are wrapped in `platform.isIOS && experimentalDevicesEnabled`. The factory-layer eligibility check is also platform-gated, so Android users get zero behavior change. Existing secondary-mode Webhook behavior (on either platform) is untouched.

**Default on upgrade:** `PrimaryMode.IndexAgent`, `shareWithIndexAgent = false`. Existing users see no UI or behavior change until they explicitly opt in.

## Changes

- `Preferences.kt`: new `PrimaryMode` enum, two new `StateFlow`s and setters, alongside existing `SecondaryMode` pattern.
- `IndexWebhookPreferences.kt`: new `isLinked` computed property.
- `RecordingOperation.kt`: `DefaultRecordingOperation` gains a `runAgent: Boolean = true` param. When false, the MCP session setup and agent invocation are skipped; transcription and entry persistence are unchanged.
- `RecordingOperationFactory.kt`: primary-gesture dispatch extracted into `createPrimaryOperation` with a four-guard eligibility check.
- `SettingsViewModel.kt`: mirrors existing `secondaryMode` exposure for the new prefs.
- `IndexSettings.kt`: new iOS-only row, new `PrimaryModeDialog` composable with Index Agent / Webhook radios, conditional Share-with-Index-agent toggle with info tooltip.

No Room migrations, no new dependencies, no new files.

## Test plan

- [ ] On iOS with `experimentalDevices = true`: Settings shows the new "Single click and hold" row above "Double click and hold". Subtitle reads "Index Agent" by default.
- [ ] Without a webhook configured, the Webhook radio is disabled and shows "Configure Webhook" button that opens the existing webhook dialog.
- [ ] With a webhook configured, selecting Webhook enables the radio; the Share-with-Index-agent toggle appears below, default OFF.
- [ ] Single click and hold gesture with Webhook + Share OFF: transcription fires webhook, no note created, recording entry still appears in feed with transcription.
- [ ] Single click and hold gesture with Webhook + Share ON: transcription fires webhook AND agent runs (notes may be created as before).
- [ ] iOS without `experimentalDevices`: "Single click and hold" row is not visible. No behavior change.
- [ ] Android with any combination: "Single click and hold" row is not visible. No behavior change.

## Notes

No tests added. Existing test suite is Android-only and would not validate iOS-specific behavior. Happy to add `commonTest` tests for the `PrimaryMode.fromId` fallback and `Preferences` round-trip if preferred.

### DIFF
--- a/experimental/src/commonMain/kotlin/coredevices/ring/database/Preferences.kt
+++ b/experimental/src/commonMain/kotlin/coredevices/ring/database/Preferences.kt
@@ -25,6 +25,8 @@ interface Preferences {
     val debugDetailsEnabled: StateFlow<Boolean>
     val approvedBeeperContacts: StateFlow<List<ApprovedBeeperContact>>
     val secondaryMode: StateFlow<SecondaryMode>
+    val primaryMode: StateFlow<PrimaryMode>
+    val shareWithIndexAgent: StateFlow<Boolean>
     val reminderProvider: StateFlow<ReminderProvider>
     val noteProvider: StateFlow<NoteProvider>
     val noteShortcut: StateFlow<NoteShortcutType>
@@ -41,6 +43,8 @@ interface Preferences {
     fun setDebugDetailsEnabled(enabled: Boolean)
     suspend fun setApprovedBeeperContacts(contacts: List<ApprovedBeeperContact>?)
     fun setSecondaryMode(mode: SecondaryMode)
+    fun setPrimaryMode(mode: PrimaryMode)
+    fun setShareWithIndexAgent(value: Boolean)
     fun setReminderProvider(provider: ReminderProvider)
     fun setNoteProvider(provider: NoteProvider)
     fun setNoteShortcut(shortcut: NoteShortcutType)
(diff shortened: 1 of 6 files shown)


derek-knox:
@iFixRobots Any particular reason you closed this? I was considering my own PR before I found this one.

PR #157, opened 2026-04-19
Add sync and build apk workflow

1c3Panda:


1c3Panda:
Apologies, sent this to the wrong place

PR #162, opened 2026-04-23
Android: Enable VoIP calls for InCallService

leo60228:
This allows Google Voice VoIP ("self-managed") calls to appear on the watch. The button to answer a call isn't present, which I think may be a limitation on the Google Voice side.


leo60228:
I missed that the DCO and CLA requirements were separate. For corroboration, I've [contributed to Linux](https://github.com/torvalds/linux/commit/532743944324a873bbaf8620fcabcd0e69e30c36) under this name.

sjp4:
Thanks! Did you verify that the fix works?

leo60228:
Yes, I did.

sjp4 (approved):

PR #190, opened 2026-05-15
Add Rebble ASR backend and STT router

jplexer:
Introduce RebbleAsrService and RebbleAsrTranscription plus STTRouter to dispatch between Cactus and Rebble ASR. Add voice config to BootConfig, new Rebble STT modes, and adjust CactusTranscriptionService to ignore Rebble modes. Wire services into watchModule and update UI imports.

Rebble's ASR is only shown as an option if the user logged in while having a rebble subscription


sjp4:
@crc-32

crc-32 (approved):

Issue #184, opened 2026-05-12
Adjust behavior when built without certain secrets


aveao:
Someone brought up the idea of this app on f-droid again and I don't personally care suuper much but it'd still be nice to make the app nicer to build from source in case someone gets the energy and time to fight with them about app inclusion.

Anyhow, I think it'd be nice for all people building from source to:
- Default to local transcription if no whisprflow (or the new thing) ~key is supplied
- Hide the relevant login buttons if the client secrets aren't supplied

That ought to make things nicer for people building from source and not wanting to utilize "non-free services".

Happy to shoot a PR for this if you're okay with accepting it. Also happy to break it down into 2 parts, one for login buttons, one for transcription settings :)

sjp4:
Go for it

aveao:
Will wait until #190 is merged before I do the one for transcription.

aveao:
Been thinking and I think one of the things I want to also do as part of this is adjusting the bug report system so that it still lets you do the export, but gets rid of the rest of options there, like viewing bug reports and actually filing new ones.

PR #191, opened 2026-05-15
Allow disabling login buttons

aveao:
One half of #184 

Screenshot_20260516_010700
Screenshot_20260516_010831

### DIFF
--- a/util/build.gradle.kts
+++ b/util/build.gradle.kts
@@ -197,6 +197,10 @@ fun gradleStringPropOrNull(name: String): String? {
     return prop
 }
 
+val googleAuthEnabled = (properties["googleAuthEnabled"] ?: project.properties["googleAuthEnabled"] ?: System.getenv("googleAuthEnabled")) == "true"
+val appleAuthEnabled = (properties["appleAuthEnabled"] ?: project.properties["appleAuthEnabled"] ?: System.getenv("appleAuthEnabled")) == "true"
+val githubAuthEnabled = (properties["githubAuthEnabled"] ?: project.properties["githubAuthEnabled"] ?: System.getenv("githubAuthEnabled")) == "true"
+
 buildkonfig {
     packageName = "coredevices.util"
     objectName = "CommonBuildKonfig"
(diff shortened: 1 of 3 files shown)


asyba (approved):

aveao:
(Also added a change so that settings in local.properties can override gradle.properties)

util/build.gradle.kts
```diff
@@ -212,6 +212,9 @@ buildkonfig {
         buildConfigField(FieldSpec.Type.STRING, "WISPR_AUTH_URL", gradleStringPropOrNull("wisprAuthUrl"), nullable = true)
         buildConfigField(FieldSpec.Type.STRING, "MEMFAULT_TOKEN", gradleStringPropOrNull("memfaultToken"), nullable = true)
         buildConfigField(FieldSpec.Type.STRING, "GOOGLE_CLIENT_ID", gradleStringPropOrNull("googleClientId"), nullable = true)
+        buildConfigField(FieldSpec.Type.BOOLEAN, "GOOGLE_AUTH_ENABLED", (gradleStringPropOrNull("googleAuthEnabled") == "true").toString())
```
sjp4: Could any of these be derived from other fields (e.g. from GOOGLE_CLIENT_ID being populated or not)?

util/build.gradle.kts
```diff
@@ -212,6 +212,9 @@ buildkonfig {
...
+        buildConfigField(FieldSpec.Type.BOOLEAN, "APPLE_AUTH_ENABLED", (gradleStringPropOrNull("appleAuthEnabled") == "true").toString())
+        buildConfigField(FieldSpec.Type.BOOLEAN, "GITHUB_AUTH_ENABLED", (gradleStringPropOrNull("githubAuthEnabled") == "true").toString())
```
sjp4: I think we should default these to `true`, not `null` (maybe need a `gradleStringPropOrNull` helper?). As is stands, this would break prod builds because we have a different `gradle.properties` file for those (which we can do, but I'd rather have good defaults)

util/build.gradle.kts
aveao: Only google auth can (other 2 use firebase auth), and between having 2 of them use this pattern and that one not, and this approach, I went with this. Let me know if I should change it.

util/build.gradle.kts
aveao: Sure, can do that.

On that note: This PR also changes things so that much if not all of gradle.properties can be overridden from a local.properties file instead. For my own builds, I put all my varying fields on local.properties now. Maybe that'd be of interest to your builds as well.

(that's why `gradleStringPropOrNull` is changed)

aveao: Done in https://github.com/coredevices/mobileapp/pull/191/commits/9d1771b3139331ac720921c77be601feb5544ea5, added a new helper to get boolean values and made it so that we can specify defaults of it. Not currently needed but it seemed like a useful feature to have.

asyba:
This will be also visible for the Bug Report section? since also requires Sign in.

aveao:
See here: https://github.com/coredevices/mobileapp/issues/184#issuecomment-4507024054

I intend to add separate gating for bug reporting options if not built as an official build.

Issue #78, opened 2025-12-25
Health Integration: Daily Typicals


michaelthatsit:
I haven't implemented in the initial integration to avoid making it even bigger. Right now we're only syncing 30 day averages.

PR #194, opened 2026-05-16
Implement sending typical sleep to the watch

aveao:
See https://github.com/coredevices/mobileapp/issues/78

~Draft as this builds on https://github.com/coredevices/mobileapp/pull/192, that should be merged first~

image


aveao:
Rebased, should be ready to review.

PR #196, opened 2026-05-17
Expose start/stop record actions with vibration feedback

awwaiid:
Feel free to close but to take as an idea -- I don't have an index01 yet, but I want to start pretending that I do or expose other ways to record voice. One simple thing is a [homescreen record/stop widget on android](https://github.com/awwaiid/pebble-recording-widget). This PR exposes the broadcast thingie and some vibration feedback to power it.


sjp4:
cc @crc-32 

(at a quick glance, if we do want something like this then I don't think it should be handled in PebbleService (or generally in the `composeApp` module)

crc-32:
I agree with the idea and we'll likely do something like this eventually, but disagree with the implementation and it will probably be part of future larger features. Thank you though!

PR #197, opened 2026-05-17
Add file chooser support in watch app settings on Android

jwlarocque:
On iOS WKWebView supports file input out of the box, but on Android onShowFileChooser does nothing by default. As a result there is a disparity in watch app settings capability (this disparity existed in the original Pebble app as well).  
This PR overrides onShowFileChooser in an AccompanistWebChromeClient passed to the WebView's platformWebViewParams in order to launch the system file chooser.  On iOS it passes null, the default, to retain current behavior.

Tested on Pixel 7 running Android 16 (with a watch app that uses \<input type="file"\> in a custom Clay component). I do not have a Mac/iOS device to build and test the iOS version; should be no change but if anyone with a Mac is able to confirm that would be fantastic.


pebble/src/androidMain/kotlin/coredevices/pebble/ui/WatchappSettingsScreen.android.kt
```diff
@@ -45,6 +57,38 @@ internal actual suspend fun restoreLocalStorage(webView: NativeWebView) {
...
+    return remember(launcher) {
+        PlatformWebViewParams(
+            chromeClient = object : AccompanistWebChromeClient() {
+                override fun onShowFileChooser(
+                    webView: WebView?,
+                    filePathCallback: ValueCallback<Array<Uri>>?,
+                    fileChooserParams: FileChooserParams?
+                ): Boolean {
+                    pending.getAndSet(filePathCallback)?.onReceiveValue(null)
+                    val intent = fileChooserParams?.createIntent() ?: return false
```
sjp4, line 77: Does this need to `pending.getAndSet(null)?.onReceiveValue(null)` before returning?

tomrav:
If there's any way of getting this PR merged, I would love to assist, this is a blocker for an app I'm making.

tomrav:
Just want to say thanks for merging this, my first watch face is now live on the store! :D

https://apps.repebble.com/c1609136e0a3430bb1bf52b1

ericmigi:
congrats, that looks great!

Issue #198, opened 2026-05-18
Apps shouldn't be able to access certain APIs on phone JS side (like audio)


aveao:
This app (despite its vibes) works by utilizing the JS audio API (works at least on android).

Cool idea, but apps being able to play arbitrary sounds on your phone is probably unpreferable, potentially some sort of permission flag per-app would be best, or just blocking out this API entirely (and shipping a native find my phone feature, which could also bypass the phone volume being turned down).

https://apps.repebble.com/find-my-phone_695da4c3fbfdde0009f99799

sjp4:
We're planning on both of these (working on a generic permission system which could be used for something like this, but also intend to block using that API once we have a built-in find-my-phone app)

saltedlolly:
This particular app is also non-functional on iOS, despite claiming otherwise, and is an example of an app that should be hidden for iOS users as discussed in #207 

That said, in this case that may not be necessary if access to the API is going to be turned off for Android users as well.

PR #199, opened 2026-05-19
iOS find phone

asyba:
Explanation here: https://github.com/coredevices/PebbleOS/pull/1326


asyba:
cc https://github.com/coredevices/mobileapp/issues/198

sjp4:
As mentioned in the FW PR, we don't want to do this without a permission system. We also want to add a native find-my-phone app which works cross-platform (handled entirely by the mobile app, not pkjs), which would would mean this isn't required (i'm not sure how many legitimate use-cases there are for your phone making a noise apart from find-my-phone)

asyba:
@sjp4 I did like that to offer if other developers wants to create there own findby app, but if core team is doing that, that is okay.

asyba:
I will try the approach as native app.

asyba:
@sjp4 with this new approach as native app?  We don't need to have a"permission system", or you still want to have it even if the feature is only used by PebbleOS stock apps?
I did't yet push the change for pebbleOS PR, depending of you anwer above.

asyba:
@sjp4 don't forget me

sjp4:
> @sjp4 with this new approach as native app? We don't need to have a"permission system", or you still want to have it even if the feature is only used by PebbleOS stock apps?

I think that's OK - rather than a permission we could have a LibPebbleConfig setting to allow users to disable it.

Using AppMessage for this or not is a good question (alternative being a new pebble protocol endpoint, if it's really a system app sending this). AppMessage has the benefit of being able to install the app that controls this on older watches. Drawback that any app could technically use this UUID to trigger the sounds (but wouldn't be allowed on the app store, so not a big issue IRL). This looks pretty sensible (without reviewing the code deeply)

PR #201, opened 2026-05-19
Extended DataLogging API for third-party Pebble watch apps

Fauxmage:
### Added
 - CustomDataLogging interface in LibPebble.kt with two data transmission mechanisms:
   - customData: SharedFlow (DROP_OLDEST).
   - setDataSink and CustomDataLoggingSink: functional interface with a suspending onData(event) for backpressured delivery.
 - Internal bounded channel (BufferOverflow.SUSPEND) and a worker coroutine in Datalogging which drains events into registered sink.

---
 
### Changes
 - Datalogging now implements CustomDataLogging, routing refactored (early-return guards for health and system-app paths, otherwise unchanged).
 - logData is now suspending: Submits external events to the sink channel while still offering them to the customData flow.
 - Datalogging constructor now takes the library coroutine scope (LibPebbleCoroutineScope), in LibPebbleModule.kt it is bound to the CustomDataLogging interface.
 - LibPebble3 accepts and delegates CustomDataLogging, exposing both the mechanisms via the LibPebble interface.


ericmigi:
Do you have a project in mind that you are planning to use this with?

UlrikBro:
> Do you have a project in mind that you are planning to use this with?

Hello, yes I am using this implementation for collecting accelerometer data, however if my changes causes any issues then I will continue using my fork until someone else implements an extension for the DataLogging API.

sjp4:
> > Do you have a project in mind that you are planning to use this with?
> 
> Hello, yes I am using this implementation for collecting accelerometer data, however if my changes causes any issues then I will continue using my fork until someone else implements an extension for the DataLogging API.

It sounds like what we really need is to implement a 3rd-party datalogging API in PebbleKit 2.0 (assuming you are using android?) - that's what used to exist in the original app, but it hadn't be recreated in the new app yet.

UlrikBro:
> > > Do you have a project in mind that you are planning to use this with?
> > 
> > 
> > Hello, yes I am using this implementation for collecting accelerometer data, however if my changes causes any issues then I will continue using my fork until someone else implements an extension for the DataLogging API.
> 
> It sounds like what we really need is to implement a 3rd-party datalogging API in PebbleKit 2.0 (assuming you are using android?) - that's what used to exist in the original app, but it hadn't be recreated in the new app yet.

If no one is currently working on an implementation of it then I can look into extending the API in PebbleKit 2.0, and yes I should probably have clearified this, I have only been using and testing with android devices for the code I commited, and will continue doing so.

sjp4:
I don't think anyone is working on that - go for it!

PR #206, opened 2026-05-20
Skip Google Play in-app update check for non-Google Play installs

aveao:
Nicety for #184. It's harmless but no point in checking and no point in resulting scary logs if it isn't installed through gplay.


sjp4 (commented):
Do you think there's any risk this doesn't work? The code looks sensible, but unfortunately it's an epic pain to test this flow (creating a dedicated build, sending that to the play store on a special internal channel, installing it, publishing an update, etc, etc) so I'm kinda scared to touch it.

aveao:
What if we first merge a debug line that just logs what it's installed through, then a release or two later we merge this?

sjp4:
That should work

aveao:
Opened #210 for that. This PR now builds on that, so it's marked as draft until that's merged, feel free to undraft later.

Issue #178, opened 2026-05-07
Quick Launch Settings shows erroneous "backlight" duplicate


burntcookie90:
![image](https://github.com/user-attachments/assets/ff23cc50-fac1-40ca-84f1-c1151db39a72)

The first one toggles backlight the second toggles "Motion Backlight"

asyba:
its fixed

PR #213, opened 2026-05-24
Fix Quick Launch label Backlight

asyba:
Fix for https://github.com/coredevices/mobileapp/issues/178


sjp4:
This already got fixed

Issue #170, opened 2026-05-02
[Feature Request] Send mobile notification when Pebble is fully charged


saltedlolly:
When you put your Pebble onto charge, and it is paired with your phone, it would be great if the Pebble mobile app could send a notification once it is fully charged. This way if you are on your phone, and your watch is charging on the other side of the room, you will be notified that your Pebble is charged and you can put it back on. Obviously it requires the Pebble to be in range of the phone but it would still be useful.

asyba:
https://github.com/coredevices/mobileapp/pull/216

saltedlolly:
@asyba Nice! Thanks for adding it. It will be a nice feature to have.

PR #216, opened 2026-05-25
Send mobile notification when watch is fully charged

asyba:
Requested https://github.com/coredevices/mobileapp/issues/170

Tested on iOS not Android.

image

### DIFF
--- a/pebble/src/commonMain/kotlin/coredevices/pebble/PebbleAppDelegate.kt
+++ b/pebble/src/commonMain/kotlin/coredevices/pebble/PebbleAppDelegate.kt
@@ -9,6 +9,7 @@ import coredevices.database.WeatherLocationDao
 import coredevices.database.insertDefaultWeatherLocationOnce
 import coredevices.firestore.UsersDao
 import coredevices.pebble.firmware.FirmwareUpdateUiTracker
+import coredevices.pebble.firmware.postWatchFullyChargedNotification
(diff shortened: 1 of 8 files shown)


pebble/src/commonMain/kotlin/coredevices/pebble/PebbleAppDelegate.kt
```diff
@@ -159,6 +163,25 @@ class PebbleAppDelegate(
                     }
                 }
             }
+            GlobalScope.launch {
+                val batteryLevels = mutableMapOf<PebbleIdentifier, Int?>()
+                val notified100 = mutableMapOf<PebbleIdentifier, Boolean>()
+                libPebble.watches.collect { watches ->
```
sjp4: This should be behind a setting in CoreConfig

The-Noah:
It would be nice if the level could be customized, so I could get notified when it's at 80% charged. Though that could be added in a later PR, this is already going to be useful!

asyba:
@sjp4 thanks Thanks for continuing, I haven't had time lately.

ericmigi:
@asyba thanks so much for all your contributions! We'd love to send you a watch and some swag as a small token of appreciation. Can you email appstore@repebble.com with your contact info?

asyba:
@ericmigi thanks so much! 
(I sent an email yesterday)

PR #226, opened 2026-05-31
Add fake watch setting and improve FakeLibPebble

jmsunseri:
Allow selecting a fake watch type from debug settings to use FakeLibPebble instead of a real Bluetooth connection. Simplify FakeLibPebble by removing random data generation and adding displayName to WatchHardwarePlatform.

image

image

image

image


jmsunseri:
cc @ericmigi

sjp4:
Why do you want this?

jmsunseri:
> Why do you want this?

I was working on a feature that required access to the settings of a watch but I was having a lot of difficulty connecting a watch to a google phone emulator and this seemed like a very simple work around so that you could access and test features that required a connected watch without the watch really being there.

jmsunseri:
@sjp4 is there anything blocking this from going forward?  i've got several things i'd like to work on and improve and this would make it so much easier to work on these things?

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/metadata/WatchHardwarePlatform.kt
```diff
@@ -14,40 +14,40 @@ import kotlinx.serialization.descriptors.SerialDescriptor
...
-    CORE_GETAFIX_DVT(20u, WatchType.GABBRO, "getafix_dvt"),
-    CORE_GETAFIX_DVT2(21u, WatchType.GABBRO, "getafix_dvt2"),
-    PEBBLE_SILK_BIGBOARD(250u, WatchType.DIORITE, "silk_bb"),
-    PEBBLE_SILK_BIGBOARD_2_PLUS(248u, WatchType.DIORITE, "silk_bb2"),
-    PEBBLE_ROBERT_EVT(13u, WatchType.EMERY, "robert_evt"),
-    PEBBLE_ROBERT_BIGBOARD(249u, WatchType.EMERY, "robert_bb"),
-    PEBBLE_ROBERT_BIGBOARD_2(247u, WatchType.EMERY, "robert_bb2"),
-    CORE_OBELIX_BIGBOARD(244u, WatchType.EMERY, "obelix_bb"),
-    CORE_OBELIX_BIGBOARD_2(243u, WatchType.EMERY, "obelix_bb2"),
+enum class WatchHardwarePlatform(val protocolNumber: UByte, private val _watchType: WatchType, val revision: String, val displayName: String, val isConsumer: Boolean = false) {
```
sjp4: Adding a user-facing name to a board doesn't really make sense - this is probably the wrong layer to do this (also many of the isConsumer flags here are wrong)

util/src/commonMain/kotlin/coredevices/util/CoreConfig.kt
```diff
@@ -93,7 +99,15 @@ data class CoreConfig(
     val obfuscateSensitiveLogs: Boolean = true,
     val hidePermissionWarningBadges: Boolean = false,
     val androidForegroundServiceForWatchConnection: Boolean = false,
-)
+    val fakeWatches: Set<WatchHardwarePlatform> = emptySet(),
+    val fakeActiveWatch: WatchHardwarePlatform? = null,
+) {
+    init {
+        require(fakeActiveWatch == null || fakeActiveWatch in fakeWatches) {
```
sjp4: this would crash the app on start

sjp4 (commented):
> @sjp4 is there anything blocking this from going forward? i've got several things i'd like to work on and improve and this would make it so much easier to work on these things?

At the base level, this breaks the only thing we currently use FakeLibPebble for (compose previews) by removing randomisation.

Given that FakeLibPebble is only designed to power a few specific screens right now, the behaviour in a lot of the app will be undefined (likely crashes, weird behaviour in many places). It's also losing a lot of functionality that is not bluetooth related (i.e. that might be better served with a mock device transport in libpebble).

If we did go this route (powering the entire app with FakeLibPebble), we do not want to expose it to users (even behind the debug options flag) - it would need to be behind a buildconfig flag configured in gradle properties. We would also not want it controlled like that in settings/libpebbleconfig - it should use the discovery flow to add new devices, and persist them internally to FakeLibPebble.

jmsunseri:
Thank you for the reply.  let me see if i can implement this in a different
manner!

Sincerely,
Justin Sunseri

jmsunseri:
@sjp4 hopefully you like this implementation better

PR #230, opened 2026-06-02
libpebble3: pure-Kotlin iOS location, drop LibPebbleSwift framework

SalvatoreT:
Reimplement IOSLocation in pure Kotlin/Native against platform.CoreLocation (a system framework auto-linked by Kotlin/Native) instead of a hand-built Swift framework. This means iOS consumers of libpebble3 no longer need a separate LibPebbleSwift framework.

- Add util/IOSLocation.kt (CoreLocation-backed); wire it into IosSystemGeolocation.
- Delete the libpebble-swift Xcode project, the libPebbleSwift cinterop/.def, the buildFrameworkLibPebbleSwift task and BuildSwiftFramework class, and the -framework LibPebbleSwift linkerOpts in libpebble3 and composeApp.


crc-32:
this sounds like a great change and yes the LibPebbleSwift framework is clunky and a hack, I originally added it because there was issues with kotlin <-> objc interop but I don't recall specifics. Does this code consistently work for you @SalvatoreT ?

sjp4:
Tested on a couple of devices locally - all looks good; thanks!

PR #238, opened 2026-06-05
libpebble3: Defer BLE GATT server pre-registration until a BLE watch is paired

bentemple:
## Description

This PR Changes the app to defer BLE GATT registration until after there is a registered pebble watch in the pebble app DB. Registration will already automatically happen as part of the pairing flow, so changing this lets both Gadgetbridge and the Pebble Mobileapp happily co-exist and the user is free to choose whichever app to use each wearable with. 

## Context
I noticed that I can't use Gadgetbridge for a BLE pebble watch, but the pebble mobile app for the Index ring simultaneously. I know this is may be an edge-case for most users, but it's impacting me.

I also saw the following review on Google play:

> When installing Pebble it forces me to uninstall another open hardware app (Gadgetbridge), which I use to connect my Casio watches. That's worse than what Casio, Amazfit and other smart device manufacturers are doing and I am shocked to see Pebble on track being evil like this.

The only conflict (that I can find) between the Pebble Mobile app and Gadgetbridge is the Pebble GATT registration of the same UUID for pebble BLE watches. Gadgetbridge registers this UUID lazily (when a Pebble watch is paired and set to auto-connect), but the pebble app _always_ registers this even if there isn't a paired watch.

## Test-Plan
Launched debug version of app
Paired pebble watch with GB
Removed all pebble watches from Pebble mobile app
Paired Index with Pebble mobile app
Verified index functionality
Verified GB could connect to the pebble watch

Removed pebble watch from GB
Paired pebble watch with pebble mobile app
Verified watch functioned normally
Verified I could force-close and re-open pebble mobile app
Pebble mobile-app auto-connected to the pebble watch as expected

### DIFF
--- a/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/bt/ble/transport/GattServer.kt
+++ b/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/connection/bt/ble/transport/GattServer.kt
@@ -8,6 +8,8 @@ import io.rebble.libpebblecommon.connection.bt.BluetoothState
 import io.rebble.libpebblecommon.connection.bt.BluetoothStateProvider
 import io.rebble.libpebblecommon.connection.bt.ble.BlePlatformConfig
 import io.rebble.libpebblecommon.connection.bt.ble.pebble.SERVER_META_RESPONSE
+import io.rebble.libpebblecommon.database.dao.KnownWatchDao
+import io.rebble.libpebblecommon.database.entity.TransportType
 import io.rebble.libpebblecommon.di.LibPebbleCoroutineScope
 import kotlinx.coroutines.channels.SendChannel
 import kotlinx.coroutines.flow.Flow
@@ -47,6 +49,7 @@ class GattServerManager(
     private val appContext: AppContext,
     private val bleConfigFlow: BleConfigFlow,
     private val blePlatformConfig: BlePlatformConfig,
+    private val knownWatchDao: KnownWatchDao,
 ) {
     private val serverMutex = Mutex()
     private val logger = Logger.withTag("GattServerManager")
@@ -57,7 +60,15 @@ class GattServerManager(
             bluetoothStateProvider.state.collect { bluetooth ->
                 logger.d("Bluetooth state: $bluetooth")
                 when (bluetooth) {
-                    BluetoothState.Enabled -> openIfNeeded()
+                    BluetoothState.Enabled -> {
+                        val hasBleWatch = knownWatchDao.knownWatches()
+                            .any { it.transportType == TransportType.BluetoothLe }
+                        if (hasBleWatch) {
+                            openIfNeeded()
+                        } else {
+                            logger.d("no BLE watches registered, skipping GATT server pre-registration")
+                        }
+                    }
                     BluetoothState.Disabled -> {
                         if (blePlatformConfig.closeGattServerWhenBtDisabled) {
                             close()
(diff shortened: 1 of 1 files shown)


sjp4:
> I noticed that I can't use Gadgetbridge for a BLE pebble watch, but the pebble mobile app for the Index ring simultaneously. 

That use-case seems a lot more tenuous than the one below... most people are using gadgetbridge for pebble because they don't want to use the official app/have the internet permission/etc, but you are using the app and still want to use gadgetbridge? :)

>I know this is may be an edge-case for most users, but I also saw the following review on Google play:
>
>When installing Pebble it forces me to uninstall another open hardware app (Gadgetbridge), which I use to connect my >Casio watches. That's worse than what Casio, Amazfit and other smart device manufacturers are doing and I am >shocked to see Pebble on track being evil like this.

This change is not going to help those users. (I know i'm preaching to the wrong person here but...) we made the choice to block them connecting if another pebble app is installed, purely for their own reliability/sanity. We have no idea if the other installed pebble app is attempting to connect to the same watch or not, and we had a *lot* of support requests initially about this (it randomly sometimes works, sometimes fails), which is why we gated it. We did add a setting to disable the check if users have that legit use-case of using gadgetbridge for a non-pebble device.

> The only conflict (that I can find) between the Pebble Mobile app and Gadgetbridge is the Pebble GATT registration of the same UUID for pebble BLE watches. Gadgetbridge registers this UUID lazily (when a Pebble watch is paired and set to auto-connect), but the pebble app always registers this even if there isn't a paired watch.

All that being said, this change seems fine.

bentemple:
> That use-case seems a lot more tenuous than the one below... most people are using gadgetbridge for pebble because they don't want to use the official app/have the internet permission/etc, but you are using the app and still want to use gadgetbridge? :)

I mean, I think when you break-down the what and the why, this makes a lot more sense (at least to me):
I think that unfettered phone notification access is one of the single most sensitive permissions an app can have (alongside always on location, phone access, text access, etc...) those encrypted signal messages? plain text. who is calling you when? where are you? etc... -- that means I have a hard-line that any app with access to that much information, cannot have internet access. These are also foundational permissions for a smartwatch to be useful, hence: Gadgetbridge.

That being said, the Index ring needs far fewer and less invasive permissions to still be useful: Set alarms, show notifications, bluetooth (and relevant find nearby devices), internet -- that's it. It's only listening when you push a button. Not to mention the pebble mobileapp _is_ open source so I already have a higher level of trust in it. So using the Index with the pebble mobile app, and the pebble watch with Gadgetbridge does make sense to me. 

Ultimately I will add support for the Index to gadgetbridge, but that's a much more long-term proposition. Probably an Index Bridge companion app that receives the audio clips via Intent, then forwarding that to a configurable MCP server for processing (locally run of course). It's just a lot more work. I want to practice using the Index now and learn how to make it a habit that can help me. Not to mention I am obligated (or at least feel a responsibility) to use the Index with the stock pebble mobile app until it launches fully. Otherwise how can I give meaningful feedback / bug reports or see what the typical user experience will be like.

> This change is not going to help those users. (I know i'm preaching to the wrong person here but...) we made the choice to block them connecting if another pebble app is installed, purely for their own reliability/sanity. We have no idea if the other installed pebble app is attempting to connect to the same watch or not, and we had a lot of support requests initially about this (it randomly sometimes works, sometimes fails), which is why we gated it. We did add a setting to disable the check if users have that legit use-case of using gadgetbridge for a non-pebble device.

yeah, I mean I actually don't even know how or why this user was having issues. There should be no conflict between pebble and gadgetbridge with respect to casio watches. But this was just what got me thinking huh, maybe I could fix this for myself (and others) :man_shrugging: Also, I will say, this pebble mobile app uses a _different_ content provider string, which means that that aspect doesn't conflict with Gadgetbridge (which uses the same as the OG pebble app). I must assume that they first paired the pebble watch with Gadgetbridge, and then tried to use it with the Pebble mobile app and were experiencing some sort of conflict? idk it doesn't make much sense to me. In my testing the Pebble mobile app always seemed to win, probably because it registers so early that it always has precedence.

> All that being said, this change seems fine.

thanks :smile:

PR #248, opened 2026-06-09
Android: trust user-installed CA certs via Network Security Config

### LINKED ISSUE #247: Android: PKJS apps can't connect to servers using user-installed CA / self-signed certs (no Network Security Config)
On Android, JS apps (PKJS) that open `https`/`wss` connections to a server using a user-installed CA or self-signed cert fail to connect, even when the user has installed and trusted that cert on their device. This is because the app trusts only the **system** CA store, not the **user** CA store, and ships no Network Security Config to opt in.

iOS is not affected in the same way (NSURLSession honors user-trusted roots once "Full Trust" is enabled), so this is an Android-specific gap.

## Repro

1. Run a server with a self-signed cert or a private CA (common with self-hosted setups, e.g. Home Assistant on a LAN).
2. Install and trust that CA on the Android device (Settings → Security → install certificate). The system browser now trusts it.
3. Install a PKJS watchapp that connects to that server over `wss://` / `https://`.
4. The connection fails at the TLS handshake. The WebSocket closes immediately (abnormal close, code 1006), and HTTP requests fail the same way.

## Cause

- Android runs PKJS inside a WebView (`libpebble3/.../js/WebViewJsRunner.kt`), so `WebSocket` / `XMLHttpRequest` go through Chromium's network stack.
- Since Android 7.0 (API 24), apps do **not** trust user-added CA certificates by default. Trusting the user store requires a Network Security Config that opts in.
- The app targets `targetSdk = 35` (`gradle/libs.versions.toml`) and declares no `networkSecurityConfig` (`composeApp/src/androidMain/AndroidManifest.xml`), and there's no `res/xml/network_security_config.xml`. So the WebView (and its WebSockets) trust system CAs only.

Net result: the cert is trusted by the device but ignored by the app, and there's nothing the user can do on-device to fix it.

## Suggested fix

Add a Network Security Config that includes the user trust store, and reference it from the manifest.

`composeApp/src/androidMain/res/xml/network_security_config.xml`:

```xml
<network-security-config>
    <base-config>
        <trust-anchors>
            <certificates src="system" />
            <certificates src="user" />
        </trust-anchors>
    </base-config>
</network-security-config>
```

Manifest `<application>` tag:

```xml
android:networkSecurityConfig="@xml/network_security_config"
```

This makes the app process (including the WebView and its `wss://` connections) honor user-installed CAs, matching what users already expect after trusting a cert on their device.

## Notes / open questions

- Worth confirming the WebView's WebSocket picks up the app NSC (Chromium on Android does consult the app config for `wss`, but good to verify on a real build).
- If trusting the user store app-wide is too broad, the same config can be scoped to specific domains via `<domain-config>`, though for PKJS the target server is user-configured and not known ahead of time, so the base-config approach is the practical option.
- Context: came up via a Home Assistant PKJS app where users self-host with self-signed certs. Original report: https://github.com/skylord123/pebble-home-assistant-ws/issues/65

skylord123:
## Problem

On Android, PKJS apps that open `https`/`wss` connections to a server using a self-signed or private-CA cert fail to connect, even after the user installs and trusts that cert on their device.

Since Android 7.0 (API 24), apps only trust the **system** CA store by default. The app targets `targetSdk = 35` and ships no Network Security Config, so the WebView that runs PKJS (and its WebSocket connections) ignores the **user** CA store. The cert is trusted by the device but ignored by the app, and there's nothing the user can do on-device to fix it.

This came up with a Home Assistant PKJS app where users self-host with self-signed certs. The WebSocket fails the TLS handshake and closes immediately (abnormal close, 1006).

## Fix

Add a Network Security Config that trusts the user store alongside the system store, and reference it from the manifest's `<application>` tag.

`composeApp/src/androidMain/res/xml/network_security_config.xml`:
```xml
<network-security-config>
    <base-config>
        <trust-anchors>
            <certificates src="system" />
            <certificates src="user" />
        </trust-anchors>
    </base-config>
</network-security-config>
```

Manifest:
```xml
android:networkSecurityConfig="@xml/network_security_config"
```

iOS already honors user-trusted roots (NSURLSession, once "Full Trust" is enabled), so this brings Android in line.

## Notes

- Closes #247

### DIFF
--- a/composeApp/src/androidMain/AndroidManifest.xml
+++ b/composeApp/src/androidMain/AndroidManifest.xml
@@ -16,6 +16,7 @@
     <application
         android:name=".MainApplication"
         android:usesCleartextTraffic="true"
+        android:networkSecurityConfig="@xml/network_security_config"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
--- a/composeApp/src/androidMain/res/xml/network_security_config.xml
+++ b/composeApp/src/androidMain/res/xml/network_security_config.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Allow user-installed CA certificates to be trusted in addition to the system
+  CA store. Since Android 7.0 (API 24), apps only trust the system store by
+  default, so PKJS apps connecting to self-hosted servers with a self-signed or
+  private-CA cert (e.g. Home Assistant over wss://) fail to connect even after
+  the user installs and trusts the cert on their device.
+-->
+<network-security-config>
+    <base-config>
+        <trust-anchors>
+            <certificates src="system" />
+            <certificates src="user" />
+        </trust-anchors>
+    </base-config>
+</network-security-config>


hoh:
This would be great!

hectorsouthern:
+1

PR #260, opened 2026-06-18
Respect phone Do Not Disturb for incoming calls

alonmln:
## Summary

Extends the existing "Respect Phone Do Not Disturb" setting so it also applies to incoming phone calls, not only regular notifications.

When Android Do Not Disturb is active and the setting is enabled, incoming calls are only forwarded to the watch if Android's notification ranking says the call is allowed by the phone's DND policy. If no matching call ranking is available in time, the call is conservatively suppressed from the watch.

## Details

- Adds call-specific DND filtering based on Android notification ranking.
- Applies the filtering in both call paths:
  - Telecom `InCallService`
  - legacy phone-state receiver fallback
- Coordinates `InCallService` ownership so the legacy receiver does not reclaim a call while DND ranking is pending.
- Correlates rankings to the current call using notification caller fields, with tolerant phone-number/name matching.
- Updates the settings copy to clarify that the setting covers notifications and calls.

## Testing

- `./gradlew :libpebble3:testDebugUnitTest --tests io.rebble.libpebblecommon.calls.CallDoNotDisturbFilterTest`
- `./gradlew testDebugUnitTest`
- `./gradlew :composeApp:assembleDebug`
- Manually installed and tested a debug build on a Samsung Galaxy S26 / Android 16.
- Verified that with "Respect Phone Do Not Disturb" enabled and phone DND active, blocked incoming calls no longer vibrate/show on the watch.
- Verified that the feature works after enabling the setting in the app.


alonmln:
Small follow-up for maintainer triage: this PR is CI-green and CLA-green, and I manually tested it on a Samsung Galaxy S26 / Android 16 with a real watch.

The behavior change is scoped to the existing "Respect Phone Do Not Disturb" setting. When the setting is off, call behavior should stay unchanged. When it is on and phone DND is active, calls are only forwarded to the watch if Android's notification ranking says the call is allowed by DND policy.

Happy to adjust the implementation if there is a preferred architecture for call/notification coordination.
cc @sjp4

alonmln:
@sjp4 Anything needed from me to advance this PR?

alonmln:
@sjp4 Is this a relevant candidate?

ericmigi:
@sjp4 bump

PR #265, opened 2026-06-23
Day/night backlight color

nick1udwig:
## Problem

Would be nice if watches with RGB backlight could automatically switch backlight color, with one color for day and one for night. 

See https://github.com/coredevices/PebbleOS/issues/1139.

## Solution

For watches that have RGB backlight, add `default_color_night` state along with time tracking for sunset/sunrise.

## Notes

1. Sister PR manages displays color based on day/night and manages changing based on sunset/sunrise: https://github.com/coredevices/PebbleOS/pull/1596
2. I considered allowing user to set sunrise/set. Instead I decided to have it default to 6am/6pm and have the mobileapp code here supply more accurate times (based on location) if available. The light color logic lives in https://github.com/coredevices/PebbleOS/pull/1596
3. UI changes: in Settings > Watch > Display under `Backlight Color` there is (for supported watches):
   - A checkbox `Change Backlight Color at Night`; when checked, show two `Color`s: `Day Color` and `Night Color`


machinaeZER0:
Per your sunrise/sunset comment - any way this could be tied to the timeline pins the watch is already generating for this? If so, then there would be no need to set manually (this might be what you're doing already!)

sjp4:
This is way too complicated - the watch already knows sunrise/sunset times via the weather blobdb. This only needs to be literally adding two prefs (enabled/color). Also this doesn't need a capability bit - the watch will ignore unknown pref keys.

nick1udwig:
> This is way too complicated - the watch already knows sunrise/sunset times via the weather blobdb. This only needs to be literally adding two prefs (enabled/color). Also this doesn't need a capability bit - the watch will ignore unknown pref keys.

Thanks for the review, will take another crack at it today or tomorrow.

nick1udwig:
I spent some time thinking about your review and also the feedback I got over on https://github.com/coredevices/PebbleOS/pull/1596, specifically the desire to allow user configuration of when to start/end night ("sunset and sunrise"). Changes in c2ad6c6

Currently, sunrise/sunset from the weather API is communicated to timeline via timelinedb, not weatherdb. And although we could, in principle, read out those pins for the original scope (where user cannot configure night start/stop), that won't work well for a user-configurable case: we need to be able to send the times over whether they're sourced from the weather API or from the user setting them manually.

Hence, I ripped out the (overcomplicated, as you said) solar calculation code and replaced it with code to grab the times from the weather API -- if the user has so configured. There are also some UI changes to allow user to choose to set by weather API or manually. Screenshots attached.

I left in the prefs for sunrise/sunset times, since:
1. They'd be required for user-input transition times, so seem like the correct way to do it
2. Scope crept and so wanted to implement the user-input transition times immediately (and did so)

The other thing worth calling out here is that I left in the cap bit. The reason is that the UI should not display the options for night time color setting if the watch doesn't have firmware to support it. The only other way I could figure to do this would be to have a condition around the firmware version i.e.
```
if (version < preDayNightVersion) {
  // render UI with day only
} else {
  // render UI with day/night
}
```
Open to alternatives here.

signal-2026-06-26-14-16-04-149_002
signal-2026-06-26-14-16-04-149_003
signal-2026-06-26-14-16-04-149_004

nick1udwig:
@sjp4 PTAL again when you have time

libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/WatchPrefEntity.kt
```diff
@@ -68,6 +69,11 @@ data class WatchPrefItem(
             logger.w { "Don't know how to encode watch pref key: $id" }
             return null
         }
+        if (type.isBacklightColorScheduleHelperPref() &&
+            !params.capabilities.contains(ProtocolCapsFlag.SupportsBacklightColorSchedule)) {
```
sjp4: We don't want capability bits governing this - there are already lots of prefs which only apply to certain firmware versions or watch models and we haven't gated them like this. We have something else in the works to hopefully improve that, but it's just extra complexity here.

pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchPrefsScreen.kt
```diff
@@ -35,28 +45,65 @@ import io.rebble.libpebblecommon.database.entity.QuicklaunchWatchPref
...
 import kotlin.uuid.Uuid
 
 // Snap the notification timeout slider to 30-second increments (20 stops across 0..600s)
 // so the displayed MM:SS value is always a clean :00 or :30.
 private const val NOTIFICATION_TIMEOUT_STEP_COUNT = 19
 
 @Composable
-fun watchPrefs(): List<SettingsItem> {
+fun watchPrefs(
+    hasDayNightBacklightSupport: Boolean,
```
sjp4: these shouldn't be needed

pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchPrefsScreen.kt
```diff
@@ -35,28 +45,65 @@ import io.rebble.libpebblecommon.database.entity.QuicklaunchWatchPref
...
+        quickLaunchOptions,
+        hasDayNightBacklightSupport,
+        weatherScheduleAvailable,
+        useWeatherSchedule,
+        onUseWeatherScheduleChanged,
+    ) {
+        val settingsByPref = settings.associateBy { it.pref }
+        val manualTimezone = preference(settingsByPref, BoolWatchPref.TimezoneSourceIsManual).valueOrDefault()
+        val weatherBacklightScheduleAvailable = weatherScheduleAvailable && !manualTimezone
+        settings.mapNotNull { item ->
```
sjp4: Don't make all those changes - just add the new pref

sjp4:
- I get the argument about putting the sunrise/sunset times in prefs because user might want to configure them manually - but that seems to break when they are using a manual timezone? I'm not completely clear if those prefs are UTC or local time?
- Adding a bunch of related features (making watch prefs dynamically hide-able, introducing the concept of a composite watch pref in the UI, making watch prefs dependent on a capability, adding params specific to this to `watchPrefs`)
- There are controls now that you can re-use for the time selection, which weren't there when this was written (see quiet time schedule prefs).

Generally, this seems to change lots of things it doesn't really need to.

PR #280, opened 2026-07-03
Add Arabic (ar_SA) language pack

kaluaim:
## Summary

- Adds an Arabic (`ar_SA`) language pack entry to `LanguagePacksJson`
- The pack is built from the [Vazirmatn](https://fonts.google.com/specimen/Vazirmatn) font (SIL OFL 1.1), chosen as the default Arabic font in the discussion on coredevices/PebbleOS#1622
- `hardware: null` so it applies to all supported platforms
- `.pbl` hosted as a release asset on my PebbleOS fork: https://github.com/kaluaim/PebbleOS/releases/tag/ar_SA-v1

## Background

This accompanies the Arabic shaping/RTL rendering fixes merged into PebbleOS (most recently coredevices/PebbleOS#1622), and was requested by @ericmigi in [this comment](https://github.com/coredevices/PebbleOS/pull/1622#issuecomment-4871233640). Tested on a Core Time 2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pebble/src/commonMain/kotlin/coredevices/pebble/services/LanguagePackRepository.kt
```diff
@@ -1886,71 +1886,11 @@ private val LanguagePacksJson = """
     },
     {
       "ISOLocal": "ar_SA",
-      "file": "https://github.com/user-attachments/files/28068835/ar_SA.pbl.zip",
+      "file": "https://github.com/kaluaim/PebbleOS/releases/download/ar_SA-v1/ar_SA.pbl",
```
gmarull, line 1889: we need to have an official server for langpacks, and a formal release process. this is really a hack and a potential security hole. Ok as a temporary solution.

kaluaim: Agreed, this is just to unblock Arabic users until there's an official langpack server. It follows the same hosting pattern as the existing entries, so happy to move it over whenever that lands.

gmarull: sure not your fault!

kaluaim:
Hi @ericmigi @sjp4 👋 Friendly nudge on this one — could it be merged to unblock Arabic users? Happy to adjust anything if needed. Thanks!

sjp4:
> hardware: null so it applies to all supported platforms

Does this work on all hardware? (e.g. OG Pebbles). The previous json had specific platforms listed because the pack would not work on others

kaluaim:
> > hardware: null so it applies to all supported platforms
> 
> Does this work on all hardware? (e.g. OG Pebbles). The previous json had specific platforms listed because the pack would not work on others

@sjp4  I used hardware: null to match other language packs, and because the per-hardware entries never showed in the app — looks like 5d31045 fixes that (unreleased, so unverified). The pack didn't render on a Time Steel (elbbeP firmware), so I've restored the 7 hardware entries; only the URL and native العربية name change now. Sideload-tested on Core 2 Duo and Core Time 2; in-app download still unverified.

PR #286, opened 2026-07-06
Add settings for Android's NotificationListener hints

puckipedia:
This allows for the Pebble's NotificationListener to "mute" notifications on the device while it is active; allowing the watch to take over the duty of vibrating for incoming notifications.

(libpebble3 already had this wired up, just needed to expose the settings :))


ericmigi:
that's pretty cool! have you tested it? How well does it work?

puckipedia:
I mute phone notifications, which works really well; but I haven't tested phone calls as much. I'm currently running a patched version of the app because I can't really live without it, and before this I used various Android Wear watches that use this flag as well. Once the watch disconnects (e.g. during a shower) it falls back to vibrating my phone, which is what I personally want :p

puckipedia:
I also haven't tested the non-compation device manager flow; I have to reason to suspect it wouldn't work, but caveat emptor

PR #292, opened 2026-07-13
DataLogging: share third-party sessions with companion apps

neelts:
# TLDR

Hi!
I've been developing a custom swimming tracker app and found that the DataLogging API doesn't allow third-party apps to read it. Could we restore the legacy API?
I've also tried using AppMessage for delivery, but unfortunately it's unreliable and requires the phone to be near the pool. Also, if the signal is lost over a certain distance, it can't recover the ACK loop, which renders it stuck forever.

# Share third-party DataLogging with companion apps

A third-party watch app can write to DataLogging (the flash spool that buffers sensor data offline and syncs when the app reconnects) but can never get its data back. The `Datalogging.logData()` function only routes health tags and telemetry, and silently drops everything else. So, you can't log offline if your phone is out of range. This means that apps have to stream live over AppMessage (your phone has to stay within Bluetooth range). This makes apps unusable for things like a swim tracker.

This brings back the old 'PebbleDataLogReceiver' path. When the app is reconnected, it starts logging data again. The health/system paths haven't been changed!

## Change

`Datalogging` + `DataLoggingService`

- Expose non-health, non-system sessions instead of dropping them: `thirdPartyData: SharedFlow<ThirdPartyDatalogRecord>`, plus `thirdPartyFinished` on session close.
- Thread the `OpenSession` timestamp (watch session start epoch) through `DataLoggingSession` so companions get a stable session id.

`PebbleKitClassic`

- Broadcast to registered companion apps, mirroring the existing `app.*` delivery:
  `com.getpebble.action.datalogging.RECEIVE_DATA` (extras: `uuid`, `data_log_timestamp`, etc) and `FINISH_SESSION`

## Testing

On Pebble Time 2 (fw v4.19.2): A watch app logged 166-byte records under a non-system UUID when the phone was out of Bluetooth range for the whole session. When the Bluetooth connection was re-established, the companion received every record (accelerometer, heart rate and compass, with no loss of data), keyed by the start of the session and finalized by 'FINISH_SESSION'. Health, Memfault or Analytics paths haven't changed.

### DIFF
--- a/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/classic/io/rebble/libpebblecommon/pebblekit/classic/PebbleKitClassic.kt
+++ b/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/classic/io/rebble/libpebblecommon/pebblekit/classic/PebbleKitClassic.kt
@@ -6,6 +6,7 @@ import android.content.IntentFilter
 import co.touchlab.kermit.Logger
 import io.rebble.libpebblecommon.connection.CompanionApp
 import io.rebble.libpebblecommon.connection.ConnectedPebble
+import io.rebble.libpebblecommon.datalogging.Datalogging
 import io.rebble.libpebblecommon.di.ConnectionCoroutineScope
 import io.rebble.libpebblecommon.di.LibPebbleKoinComponent
 import io.rebble.libpebblecommon.js.CompanionAppDevice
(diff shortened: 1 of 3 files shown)


sjp4:
This should go through PebbleKit-Android 2. https://github.com/pebble-dev/PebbleKitAndroid2/issues/1 might be a good place to discuss (but it should likely use the existing bound service architecture, rather than Intents like original PebbleKit, which are much more limited in modern android)

PR #295, opened 2026-07-15
Add AMOLED-style Black theme option

ForeverFreed:
# Add AMOLED black theme

Adds a third app theme — **Black** — alongside Light and Dark, for OLED screens where a true-black background costs no backlight power.
Also fixes some minor readability issues with the original Dark theme, which would have carried over to the new Black theme (see below).

## What you get

A new "Black" option in the app theme picker. It's the existing Dark scheme with `background`, `surface`, and the lowest container role painted pure black.

## The interesting part: the surface ladder

Material 3 gives every scheme a ladder of container colours (`surfaceContainerLowest` → `surfaceContainerHighest`) that components use to lift themselves off the page. A `DropdownMenu`, for example, paints itself `surfaceContainer`.

Both dark schemes had a broken ladder when I first implemented this:

- `surfaceContainer` was set to the **same colour as `surface`**, so popups had almost no visible edge against the page behind them.
- In the Dark scheme, `surfaceContainerHighest` was **darker** than the page above it, so elevation read inverted.
- Everything left unnamed fell through to the M3 dark defaults, which are near-black — wrong against a `#333333` page.

Both ladders now ascend cleanly. Because the scheme carries the difference, components separate themselves correctly with **no per-component styling** — which is why this PR deletes the `== Grey` menu special case in `WatchesScreen` rather than generalising it.

Light theme is deliberately untouched: it separates menus via shadow elevation, which works on white. Dark themes suppress shadows, which is exactly why M3 uses tonal elevation there instead.

## Theme-carried `isDark`

`CoreAppColorScheme` now carries an `isDark` flag instead of callers testing `== Grey`. Two places were picking a dark asset by comparing against the Dark theme specifically, so Black got the wrong one: the Wispr Flow logo (rendered black-on-black) and the webview's `prefers-color-scheme`. Both now follow the flag.

## Files

10 files, +121 / −51. The substance is in `util/src/commonMain/kotlin/theme/Theme.kt`; most of the rest is one- to three-line touches.

## Testing

- `:composeApp:assembleDebug` — passing
- `:util:testDebugUnitTest` — passing, including new coverage asserting both dark ladders ascend and that menu containers clear the page colour

**Not yet verified on iOS.** The only iOS-side change is one deleted import in `PebbleWebview.ios.kt` (`isDark` became a member property, so it resolves without one). It could not be compiled locally: `:util:compileKotlinIosSimulatorArm64` fails in `libpebble3` on an unresolved `LibPebbleSwift` reference, which needs `podInstall` and reproduces on a clean checkout — pre-existing, unrelated to this change. Worth a build on macOS before merge.

## Worth a look when reviewing

The Dark theme ladder fix **restyles the shipped Dark theme**, not just the new Black one. It was unavoidable — the ladder is what makes menus work without per-component styling, so leaving Dark broken would have meant keeping the special case. Surfaces in Dark will look slightly different: containers now sit lighter than the page rather than darker.

## AI Disclosure

I used Claude (Sonnet 5 and Opus 4.8) to work on this feature. I am very new to Kotlin (but not coding in general), so using AI to help build out a maintainable new theme like this seemed appropriate. I am fully open to hearing out any changes you'd like to see before this is merged.

### DIFF
--- a/composeApp/src/androidMain/kotlin/coredevices/coreapp/MainActivity.kt
+++ b/composeApp/src/androidMain/kotlin/coredevices/coreapp/MainActivity.kt
@@ -102,6 +102,7 @@ class MainActivity : ComponentActivity() {
         val themeRes = when (theme) {
             CoreAppTheme.Light -> light
             CoreAppTheme.Dark -> dark
+            CoreAppTheme.Black -> R.style.Theme_CoreApp_Black
             CoreAppTheme.System -> if (isNightMode(this)) {
                 dark
             } else {
--- a/composeApp/src/androidMain/res/values/themes.xml
+++ b/composeApp/src/androidMain/res/values/themes.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<resources>
+    <!-- Window background is painted before Compose composes; the platform dark theme's grey
+         would flash before the black scheme takes over. -->
+    <style name="Theme.CoreApp.Black" parent="@android:style/Theme.Material.NoActionBar">
+        <item name="android:windowBackground">@android:color/black</item>
+    </style>
+</resources>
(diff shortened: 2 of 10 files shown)


ericmigi:
thanks @ForeverFreed! Keep in mind that as things change, we might not check them on this theme. If you notice any problems, please feel free to send another PR or mention the problem here.

ForeverFreed:
> thanks @ForeverFreed! Keep in mind that as things change, we might not check them on this theme. If you notice any problems, please feel free to send another PR or mention the problem here. 

Thanks @ericmigi! I finally got my PT2 just today and can't wait to continue making this project the best it can be!

PR #299, opened 2026-07-17
Enable user-defined appstore sources

ttmm-sp-zoo:
Two small changes that unlock functionality already implemented in the app, so that third-party appstore feeds can be used end to end. No refactoring.

## 1. Enable the "Add Source" / "Delete Source" UI

`AppstoreSettingsScreen.kt`

The Add Source FAB and the per-source delete button were fully implemented but commented out. `CreateAppstoreSourceDialog`, `AppstoreSettingsScreenViewModel.addSource`/`removeSource` and `AppstoreSourceDao` are already wired up, so this only uncomments the UI entry points.

Delete is deliberately offered **only for user-added sources**, guarded with the existing `isPebbleFeed()` / `isRebbleFeed()` helpers. Removing a built-in feed is unsafe today: deleting the Rebble source strands the user, as `AppstoreSourceInitializer` never re-adds it, while deleting the Pebble source trips that initializer's re-init path, which deletes and recreates every source and would silently drop user-added ones. Happy to drop this guard if you'd rather handle it differently.

The entered URL is normalized (`trim().trimEnd('/')`) to the same canonical form as the built-in feed constants. Until now every `AppstoreSource.url` came from `INITIAL_APPSTORE_SOURCES`, so it was canonical by construction; this dialog is the first path where the value is typed by a user. The URL is string-concatenated into request URLs (`"${source.url}/v1/apps/id/$id"`) and equality-compared in `addHeart()` / `isLoggedIn()`, so a pasted trailing slash would otherwise yield `.../api//v1/apps/id/...` and a source that looks correct but silently serves nothing.

## 2. Let appstore deep links specify the feed to open from

`PebbleDeepLinkHandler.kt`

`handleAppstore()` already takes the feed URL as a parameter and resolves it against the enabled sources — only the call site was fixed to `https://appstore-api.rebble.io/api`. As a result an app card could only ever be opened from Rebble.

The feed is now read from an optional `?source=` query parameter:

```
pebble://appstore/{id}?source=https%3A%2F%2Fapps.ttmm.is%2Fpebble%2Fapi
```

The `?source=` value must be URL-encoded (as above) — on iOS a raw, unencoded URL in the query does not survive URL parsing and the link will not navigate.

- **Existing links are unaffected.** With no `?source=`, the handler falls back to the Rebble feed exactly as before (the inline URL is replaced by the `REBBLE_FEED_URL` constant it duplicated).
- **Unknown or disabled feeds are handled.** The source is matched against enabled sources only; anything else resolves to no store and hits the handler's existing `"Failed to find app in enabled feeds"` message. No new failure path.

`?source=` matches the stored source URL exactly, so a link must use the same URL the source was added with — in the canonical, no-trailing-slash form the dialog now stores.

## Why

Third-party appstore feeds work today apart from this last mile. A working example is **apps.ttmm.is** — a store for TTMM watchfaces serving the full store API (per-platform descriptions, screenshots, collections). Browsing → installing a `.pbw` → loading the watchface config already works end to end on a physical device. What's missing is the ability to add the feed in the official app and to open an app card from it.

This was agreed by email with Eric Migicovsky.

## How to test

1. Appstore Sources → tap the **+** FAB → add a source, e.g. name `TTMM`, URL `https://apps.ttmm.is/pebble/api`. Adding it with a trailing slash or surrounding whitespace should behave identically.
2. The source appears with its collections and can be enabled/disabled. User-added sources can be removed with the delete button; the built-in Pebble and Rebble feeds cannot.
3. Open `pebble://appstore/{id}?source=https%3A%2F%2Fapps.ttmm.is%2Fpebble%2Fapi` for an app id served by that feed — the card opens from that source instead of failing against Rebble. Install the `.pbw` from the card and confirm it loads on the watch.
4. Regression: `pebble://appstore/{id}` with no `?source=` still opens the Rebble card while the Rebble source is enabled.
5. Error path: `?source=` pointing at a disabled or unknown feed shows "Failed to find app in enabled feeds" and does not navigate.

## Testing status — please read

**This has not been built or run.** The change is written and reviewed statically, but no JDK or Android SDK was available in the environment used, so neither the Kotlin compile, the existing test suite, nor a device run were executed. It needs a real build and an on-device pass before merging.

No tests were added. `PebbleDeepLinkHandlerTest` currently covers only the pure `parseTokenFrom` helper; exercising `handleAppstore` would need fakes for `LibPebble`, analytics and the source DAO, and writing tests I cannot execute seemed worse than saying so plainly. Happy to add coverage if you want it in this PR.

### DIFF
--- a/pebble/src/commonMain/kotlin/coredevices/pebble/PebbleDeepLinkHandler.kt
+++ b/pebble/src/commonMain/kotlin/coredevices/pebble/PebbleDeepLinkHandler.kt
@@ -8,6 +8,7 @@ import coredevices.libindex.device.IndexPlatformBluetoothAssociations
 import coredevices.libindex.device.REQUEST_URI_HOST
 import coredevices.pebble.account.PebbleAccount
 import coredevices.pebble.firmware.FirmwareUpdateUiTracker
+import coredevices.pebble.services.REBBLE_FEED_URL
 import coredevices.pebble.ui.NavBarRoute
 import coredevices.pebble.ui.PebbleNavBarRoutes
 import io.rebble.libpebblecommon.connection.AppContext
@@ -78,7 +79,10 @@ class RealPebbleDeepLinkHandler(
             uri.scheme == "pebble" -> {
                 when (uri.host) {
                     CUSTOM_BOOT_CONFIG_URL -> handleBootConfig(uri.path)
-                    STORE_URL -> handleAppstore("https://appstore-api.rebble.io/api", uri.path)
+                    STORE_URL -> handleAppstore(
+                        uri.getQueryParameter(STORE_SOURCE_PARAM) ?: REBBLE_FEED_URL,
+                        uri.path,
+                    )
                     NAVBAR_URL -> handleNavbar(uri.path)
                     REGISTER_INDEX_COMPANION_HOST -> handleRegisterIndexCompanion()
                     SHOW_WATCHES_HOST -> handleShowWatches(uri.path)
@@ -269,6 +273,7 @@ class RealPebbleDeepLinkHandler(
     companion object {
         private const val CUSTOM_BOOT_CONFIG_URL: String = "custom-boot-config-url"
         private const val STORE_URL: String = "appstore"
+        private const val STORE_SOURCE_PARAM: String = "source"
         private const val NAVBAR_URL: String = "navbar"
         private val SHOW_WATCHES_HOST = "show-watches"
 //        private val UPDATE_WATCH_NOW_HOST = "update-watch-now"
--- a/pebble/src/commonMain/kotlin/coredevices/pebble/PebbleDeepLinkHandler.kt
+++ b/pebble/src/commonMain/kotlin/coredevices/pebble/PebbleDeepLinkHandler.kt
@@ -8,6 +8,7 @@ import coredevices.libindex.device.IndexPlatformBluetoothAssociations
 import coredevices.libindex.device.REQUEST_URI_HOST
 import coredevices.pebble.account.PebbleAccount
 import coredevices.pebble.firmware.FirmwareUpdateUiTracker
+import coredevices.pebble.services.REBBLE_FEED_URL
 import coredevices.pebble.ui.NavBarRoute
 import coredevices.pebble.ui.PebbleNavBarRoutes
 import io.rebble.libpebblecommon.connection.AppContext
@@ -102,7 +103,10 @@ class RealPebbleDeepLinkHandler(
             uri.scheme == "pebble" -> {
                 when (uri.host) {
                     CUSTOM_BOOT_CONFIG_URL -> handleBootConfig(uri.path)
-                    STORE_URL -> handleAppstore("https://appstore-api.rebble.io/api", uri.path)
+                    STORE_URL -> handleAppstore(
+                        uri.getQueryParameter(STORE_SOURCE_PARAM) ?: REBBLE_FEED_URL,
+                        uri.path,
+                    )
                     NAVBAR_URL -> handleNavbar(uri.path)
                     REGISTER_INDEX_COMPANION_HOST -> handleRegisterIndexCompanion()
                     SHOW_WATCHES_HOST -> handleShowWatches(uri.path)
@@ -362,6 +366,7 @@ class RealPebbleDeepLinkHandler(
         private const val RESERVED_SIDELOAD_PREFIX = "pending_sideload_"
         private const val CUSTOM_BOOT_CONFIG_URL: String = "custom-boot-config-url"
         private const val STORE_URL: String = "appstore"
+        private const val STORE_SOURCE_PARAM: String = "source"
         private const val NAVBAR_URL: String = "navbar"
         private val SHOW_WATCHES_HOST = "show-watches"
 //        private val UPDATE_WATCH_NOW_HOST = "update-watch-now"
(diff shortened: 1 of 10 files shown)


ericmigi:
is https://apps.ttmm.is/api up? 
CleanShot 2026-07-18 at 13 59 05@2x

ttmm-sp-zoo:
I was hope that everything is fine. So it seems that Claude didn’t finished the project correctly. This need to wait 2 weeks. I just started holiday. If you can write me what is missing. Best,AlbertSent from my iPhoneOn 18 Jul 2026, at 22:59, Eric Migicovsky ***@***.***> wrote:﻿ericmigi left a comment (coredevices/mobileapp#299)
is https://apps.ttmm.is/api up?
***@***.*** (view on web)

ttmm-sp-zoo:
Hi Eric,

Back from holiday — and with a proper answer to your question in the PR. The server is up, in full. You hit a 404 because the base is https://apps.ttmm.is/pebble, not /api — all routes live under /pebble/api/v1/, mirroring appstore-api.repebble.com.

Try these: apps.ttmm.is/pebble/api/v1/home/watchfaces (12 TIC watchfaces), /home/watchapps (TTMM Utilities), or a full card at /apps/id/c18cd584537b69cc926d53f0 — TTMM Alarms 1.1.0, released today, with descriptions, screenshots and changelog. The pbw installs from apps.ttmm.is/pebble/pbw/ttmm-alarms-v1.1.0.pbw.

The PR is unchanged since July: small diff, compiles, still needs an on-device pass — but now there's a live feed to test it against.

Thanks,
Albert
TTMM · ***@***.*** ***@***.***> · ttmm.is

ericmigi:
tested

PR verdict - Ul works, but 3 findings for review:
1. Blocker: user-added sources are wiped on every app relaunch. AppstoreSourceInitializer.needsInit
includes current. any { it.algoliaAppId == null } — user sources always have null algolia fields, so every
launch deletes all sources and reinserts builtins. Watched it happen: TTMM in DB, relaunch, gone. Pre-existing
initializer code, but it makes the PR's feature session-only.
2. TTMM can't appear in store-home browse at all: AppStoreHome. applications is a required field, TTMM's
/v1/home/* returns data instead — deserialization fails silently, zero collections stored (confirmed in the on-
device DB). Add-source + refresh shows the source but no collections, and no TTMM section in the store. Needs
either app-side leniency (default for applications) or a feed-side fix.
3. Deep link needs a URL-encoded source param on iOS. Raw ?source=https://apps.ttmm.is/pebble/api (as
written in the PR description) does not navigate; ?source=https%34%2F%2F ..• opens the card perfectly. The PR
docs should say so.

ttmm-sp-zoo:
@ericmigi Thank you for the on-device pass — this is exactly the review the PR needed. All three findings were real, and fixing #2 uncovered a fourth. Status:

**1. Sources wiped on relaunch — fixed, new commit in this PR.** The initializer treated the whole sources table as builtin state: any row with `algoliaAppId == null` — i.e. every user-added source — re-triggered `needsInit`, and init deleted all rows before reinserting the builtins. The commit scopes the migration checks and the delete to the two builtin feed URLs, so user sources now survive relaunch. First-run seeding and both migration paths are unchanged for the builtin rows. (It also swaps the inline repebble URL for the existing `PEBBLE_FEED_URL` constant.)

**2. Store-home browse — fixed on the feed side, live now.** `/v1/home/*` returns `applications` (was `data`). While fixing it I checked the whole envelope against `AppStoreHome` and found a second mismatch hiding behind the one you hit: our `categories` objects were missing `application_ids`, `icon` and `links`, which `StoreCategory` requires — the rename alone would still have deserialized to zero collections. Both fixed; every required field of `AppStoreHome` / `StoreCategory` / `StoreCollection` / `StoreApplication` is verified against the live responses. So no app-side change should be needed — though if you'd like `applications` to default to empty as belt-and-braces against other third-party feeds, happy to add that here too.

**3. Deep link encoding — PR description updated.** The `?source=` value must be URL-encoded on iOS; the description now says so and carries a working example: `pebble://appstore/c18cd584537b69cc926d53f0?source=https%3A%2F%2Fapps.ttmm.is%2Fpebble%2Fapi`.

One note for re-testing browse on a Time 2: the 12 TIC watchfaces intentionally ship no emery builds (no auto-scaled faces without the author's consent), so the watchfaces browse will be thin there — TTMM Alarms is emery-only and shows under Apps. A Core 2 Duo or any basalt/diorite watch shows the full watchfaces browse.

sjp4:
Fixed a few things, and added a new deep link to add a feed (`pebble://name/url` so for this feed it would be `pebble://add-store-feed/TTMM/https%3A%2F%2Fapps.ttmm.is%2Fpebble%2Fapi`).

FYI @ttmm-sp-zoo trying to hide apps from Emery isn't supported right now. They are hidden from the collection, but a user can still install them on PT2 if they have them in the locker (and they can easily put them in the locker by connecting a different watch or overriding the watch type in the UI). Right now it looks confusing in the TTMM feed (when PT2 is connected all the faces show up on the home screen, but clicking on a collection will show blank).

ttmm-sp-zoo:
@sjp4 Thank you — for the review, the fixes, and especially `pebble://add-store-feed/{name}/{url}`: a prefilled Add Source dialog is exactly what a third-party feed needed to be usable by normal humans. The cache-keyed-by-id fix is appreciated too. Once a release with this ships, that link goes on our site as the "Add TTMM" button.

On the Emery confusion — fair catch, and it was an inconsistency in our feed, not the app. Fixed and live now: `/home`, `/apps/dev` and `/apps/category` apply the same supported-platform filter that `/apps/collection` already did. With a PT2 connected the TTMM watchfaces browse is now consistently empty instead of showing faces whose collections then come up blank; TTMM Alarms (emery-only) still shows under Apps. `/apps/id` and `/apps/bulk` stay unfiltered on purpose, so locker installs and update checks keep working for people who own the faces.

Hiding the aplite–diorite faces from PT2 is intentional on my side: I don't want auto-scaled versions of these designs offered on the new screens — PT2-native TTMM faces are coming as their own collection. I'm aware the locker / watch-type-override path can still install them; that's fine, the flag expresses intent, not DRM.

And I see it is already merged — thank you for landing it.

PR #302, opened 2026-07-19
index: split webhook into independent per-gesture configs

derek-knox:
> [!NOTE]  
> YAGNI, but I started to sense storing multiple (3+) webhook configs being valuable where UX-wise there'd be a mechanism to enable quick swapping or additional gestures (triple click & hold, etc.).

## TLDR

- Update webhooks support so each gesture (singe|double click & hold ) can be routed to a unique webhook url
- Resolved conflicts with [this PR](https://github.com/coredevices/mobileapp/pull/298) so `X-Index-Trigger` header is still used resulting in best of both worlds

|Settings|Single gesture|Double gesture|
|-|-|-|
||||

## Notes

Since the existing `SecondaryMode`'s webhook entry (legacy `id=2`) is retired ([Preferences.kt#L297-L298](https://github.com/coredevices/mobileapp/blob/1399c4b73dc993697a37ce5dcdf9f81180602405/experimental/src/commonMain/kotlin/coredevices/ring/database/Preferences.kt#L297-L298)) I opted for a path that seemed to best adhere to existing conventions.

As can be seen in the above working app screenshots, the **Index 01 > Settings > Advanced** refactors the single webhook cta into two ctas so there is now a config per gesture.

## Details

Implements the invited split-webhook feature as per-gesture webhook configs on top of the existing decorator model — the lowest-friction path that aligns with current patterns. Merged with master, so #298's `X-Index-Trigger` header is reconciled here (folded together) rather than reverting either side:

- `IndexWebhookPreferences`: independent config (URL / headers / payload mode) per gesture via new `IndexWebhookGesture` — which also owns the `X-Index-Trigger` header name + wire values (`single-click-hold` / `double-click-hold`); old `IndexWebhookTrigger` **and** #298's duplicate `IndexWebhookRecordingTrigger` removed
- Migration: legacy auth token → headers first, then flat config seeded into the gesture(s) the old trigger covered; flat keys removed
- `RecordingOperationFactory`: wraps by gesture; upload operation re-reads its gesture's config after the inner operation, so unlinking mid-processing suppresses the upload
- `IndexWebhookApi`: `uploadIfEnabled()` → `upload(config, …, gesture)`, now a stateless client that emits `X-Index-Trigger` (stripping any user-set override); unused `isEnabled` flow removed
- Header provenance keeps #298's omit-when-unknown contract: routing uses a non-null `configGesture` (single-click slot); the header uses a nullable `observedGesture`, omitted for local (phone-mic) recordings and unrecoverable legacy retries
- Settings UI: two entry points (single / double click & hold), each independently configurable; trigger radio group removed; dialog note that `X-Index-Trigger` disambiguates when both gestures share a URL
- Tests: `IndexWebhookPreferencesTest` (12 cases: migrations, round-trips, per-gesture independence) + `IndexWebhookGestureTest` locks the header wire values
- Docs: `INDEX_WEBHOOK_API.md` setup + `X-Index-Trigger` sections updated


ericmigi:
yup we'll definitely be adding this into settings in near future

derek-knox:
> yup we'll definitely be adding this into settings in near future

@ericmigi Want me to close this PR out in favor of your planned approach (assuming it differs)? If so, what's the rough planned difference in the implementation?

ericmigi:
No eta but it will definitely happen

ericmigi:
part of this may now be fixed with https://github.com/coredevices/mobileapp/pull/298 and is in latest beta app on repebble.com/app (install beta at bottom)

remainder is still on roadmap

derek-knox:
> part of this may now be fixed with #298 and is in latest beta app on repebble.com/app (install beta at bottom)
> 
> remainder is still on roadmap

👍 I updated this PR last week when that merged (above PR desc hits on it)

PR #318, opened 2026-08-06
Add watch charge limit setting

adipascu:
# Add watch charge limit setting

Adds a Charge Limit row to the Battery screen, backed by the `chargeLimitPct` watch pref introduced in coredevices/PebbleOS#1841. Capping charge below full is the usual way to slow lithium ageing, and the firmware side already enforces it. This is the phone-side control and the encoder that lets the value reach the watch at all.

| Light | Dark |
| --- | --- |
| Charge Limit in light mode | Charge Limit in dark mode |

## Row layout

The value sits in the row's leading slot rather than the trailing one, in a slot the same width as the checkbox every boolean setting already uses, so the Battery screen's rows line up with each other. `basicSettingsDropdownItem` gains an opt-in `controlOnLeft`; every other screen keeps the trailing value it has today.

## Why an encoder is the substantive part

`WatchPrefItem.value()` looks the key up in `WatchPref.from(id)` and returns null for anything it does not recognise, so an unregistered pref is dropped before it reaches the blobdb insert. Without the `EnumWatchPref.ChargeLimit` entry the app logs `Don't know how to encode watch pref key: chargeLimitPct` and the write is silently discarded, which is what the shipping build does today. Registering the key is what makes the row do anything.

## Range handling

The pref is `Off` by default, then 50% to 95% in 5% steps, carried as `TypeUInt8`. The firmware accepts `0` or `50` to `95` and validates on sync and at boot, so the app must not emit anything else. It structurally cannot: `encodeValue` is the enum's own code, and the enum only defines those values.

Decoding is the direction that needs guarding, because it consumes whatever the watch or another client wrote. Off-step values snap down to the step below, and anything outside the accepted range reads as `Off`, which matches how the firmware would treat it. `ChargeLimitPrefTest` pins that behaviour, including the case where a nonzero value below 50 must not survive as a limit.

## Visibility

A watch pref can now declare a `minFirmwareVersion`, the same idea the HRM prefs already use inline. `WatchPrefItem.value()` skips the pref below that version so it is never sent to firmware that would reject the key, and the Battery screen and the watch settings list hide the row for the same watches. Prefs without a minimum behave exactly as before.

`ChargeLimit` is pinned to `v4.37.0`, the latest firmware tag at the time of writing, so builds from `main` that carry the firmware change qualify. It should move to the release that actually ships coredevices/PebbleOS#1841 once that is known. `ChargeLimitValueTest` covers the boundary on both sides and a pref without a minimum.

---

### DIFF
--- a/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/WatchPrefEntity.kt
+++ b/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/WatchPrefEntity.kt
@@ -14,6 +14,7 @@ import io.rebble.libpebblecommon.database.dao.ValueParams
 import io.rebble.libpebblecommon.database.dao.WatchPreference
 import io.rebble.libpebblecommon.database.entity.QuickLaunchSetting.Companion.toJson
 import io.rebble.libpebblecommon.metadata.WatchType
+import io.rebble.libpebblecommon.packets.ProtocolCapsFlag
 import io.rebble.libpebblecommon.packets.blobdb.TimelineAttribute
 import io.rebble.libpebblecommon.packets.blobdb.TimelineItem.Attribute
 import io.rebble.libpebblecommon.services.blobdb.DbWrite
@@ -66,6 +67,12 @@ data class WatchPrefItem(
         val type = WatchPref.from(id)
         if (type == null) {
             logger.w { "Don't know how to encode watch pref key: $id" }
+            return null
+        }
+        if (id == EnumWatchPref.ChargeLimit.id &&
+            !params.capabilities.contains(ProtocolCapsFlag.SupportsChargeLimit)
+        ) {
+            logger.d { "Watch does not support a charge limit" }
             return null
         }
         logger.v { "trying to insert watch pref to watch blobdb: $id / $value" }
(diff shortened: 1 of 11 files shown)


libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/database/entity/WatchPrefEntity.kt
```diff
@@ -534,6 +557,23 @@ enum class EnumWatchPref(
         defaultValue = WatchLanguage.Custom,
         options = WatchLanguage.entries,
     ),
+    ChargeLimit(
+        id = "chargeLimitPct",
+        displayName = "Charge Limit",
+        description = "Limit charging to extend your battery's lifespan",
+        defaultValue = ChargeLimitLevel.Off,
+        options = ChargeLimitLevel.entries,
+        minFirmwareVersion = FirmwareVersion.sentinel(4, 37, 0),
```
sjp4, line 566: We shouldn't do this. The previous use of it was needed because we were delaing with old firmwares where we couldn't add a capability bit. (I saw on the firmware PR that Gerard commented that you shouldn't use a capability bit - I think he just meant that the firmware will ignore the pref if it's not supported).

pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchPrefsScreen.kt
```diff
@@ -157,6 +163,21 @@ fun WatchPref<*>.section(): Section = when (this) {
...
+    EnumWatchPref.ChargeLimit -> Section.Battery
+}
+
+@Composable
+internal fun chargeLimitSettingsItem(): SettingsItem? {
+    val libPebble = rememberLibPebble()
+    val config by libPebble.config.collectAsState()
+    val watches by libPebble.watches.collectAsState()
+    val firmwareVersions = remember(watches) { watches.knownFirmwareVersions() }
+    if (!config.watchConfig.enableWatchSettingsSync || !firmwareVersions.anySupports(EnumWatchPref.ChargeLimit)) {
```
sjp4, line 175: Remove this check:
- Can't get to this screen anyway without settings sync being enabled
- We've never done a version/capability check on a new watch setting before - I don't think it's needed here either (we just assume that users will update their firmware to a supported version)

pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchSettingsScreen.kt
```diff
@@ -2679,6 +2682,7 @@ fun <T> basicSettingsDropdownItem(
     show: () -> Boolean = { true },
     isDebugSetting: Boolean = false,
     extraSupportingContent: (@Composable () -> Unit)? = null,
+    controlOnLeft: Boolean = false,
```
sjp4, line 2685: I didn't understand why we need to swap left/right here (the PR description has a comment about it being aligned with other battery screen elements but I don't know what that meant). Simpler would be better

sjp4 (commented):
The screenshot has two new settings but I only see one in the code?

PR #321, opened 2026-08-06
Make firmware sideloading scriptable over adb

adipascu:
Flashing a development build meant driving the phone UI by hand, which does not fit into a build script. This adds an adb-only broadcast that takes the same sideload.

Builds on the sideload flow from coredevices/mobileapp#326.

`FirmwareSideloadReceiver` takes a `coredevices.pebble.SIDELOAD_FIRMWARE` broadcast with a `path` extra, and starts the same sideload with no confirmation, so a build script can drive it. Two things gate it: the sender must hold `android.permission.DUMP`, whose protection level is `signature|privileged|development`, which an adb shell has and an installed app cannot get, and Show debug options must be on. A broadcast arriving with debug options off now says so in a snackbar rather than only in the log.

The receiver's work runs off the main thread behind `goAsync()`, since `onReceive` would otherwise read settings and stat external storage there. Note this does not hold the process for the whole wait-for-watch window, because `sideloadFirmware` is fire and forget. Doing that properly needs the foreground service and is left alone here.

coredevices/PebbleOS#1845 documents both ways of installing a `.pbz`, from a file manager and over adb, in the firmware developer docs.

---

### DIFF
--- a/pebble/src/androidMain/AndroidManifest.xml
+++ b/pebble/src/androidMain/AndroidManifest.xml
@@ -5,6 +5,14 @@
     <application>
         <receiver android:name=".ui.TestActionReceiver" android:exported="false" />
         <receiver android:name="coredevices.pebble.firmware.UpdateActionReceiver" android:exported="false" />
+        <receiver
+            android:name="coredevices.pebble.firmware.FirmwareSideloadReceiver"
+            android:exported="true"
+            android:permission="android.permission.DUMP">
+            <intent-filter>
+                <action android:name="coredevices.pebble.SIDELOAD_FIRMWARE" />
+            </intent-filter>
+        </receiver>
 
         <provider
             android:authorities="coredevices.coreapp.pebblekit"
--- a/pebble/src/androidMain/kotlin/coredevices/pebble/firmware/FirmwareSideloadReceiver.kt
+++ b/pebble/src/androidMain/kotlin/coredevices/pebble/firmware/FirmwareSideloadReceiver.kt
@@ -1,0 +1,59 @@
+package coredevices.pebble.firmware
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import co.touchlab.kermit.Logger
+import com.russhwolf.settings.Settings
+import coredevices.pebble.PebbleDeepLinkHandler
+import coredevices.pebble.ui.showDebugOptions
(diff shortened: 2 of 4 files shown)


pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchesScreen.kt
```diff
@@ -1394,7 +1394,7 @@ fun WatchMenu(watch: PebbleDevice, navBarNav: NavBarNav) {
                 val active by watch.devConnectionActive.collectAsState()
                 val canUseDevConnection = user?.isAnonymousUser == false || config.watchConfig.lanDevConnection
                 DropdownMenuItem(
-                    text = { Text("Dev Connection") },
+                    text = { Text("SDK connection") },
```
sjp4: Don't change this (it has been called Dev Connection forever - that will confuse people)

pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchSettingsScreen.kt
```diff
@@ -1390,8 +1390,8 @@ fun rememberSettingsItemsState(navBarNav: NavBarNav?, snackbarDisplay: SnackbarD
                     },
                 ),
                 basicSettingsToggleItem(
-                    title = "Show debug options",
-                    description = "Show some extra debug options around the app - not useful for most users (contains some options which might break your watch)",
+                    title = "Developer mode",
```
sjp4: Don't change this (it's not just developer things - we'll rework this at some point but not here)

sjp4 (commented):
Looks great and useful, thanks! But please revert the text changes I mentioned

pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchesScreen.kt
adipascu: You're right, dropped the rename from this PR.

pebble/src/commonMain/kotlin/coredevices/pebble/ui/WatchSettingsScreen.kt
adipascu: Agreed, dropped as well.

PR #327, opened 2026-08-08
Index: add custom click action bindings

vishal-android-freak:
Part 1 of 3 of a stacked series adding user-configurable click actions to Index 01. Each part is a separate PR so they can be reviewed independently; please take them in order.

- **#1 (this) — click-count bindings + dispatcher + "Ask the agent"**
- #2 — "Run a tool" (MCP tool call)
- #3 — "Call webhook"

## What this does

Lets the user bind a run of 1–8 short clicks on Index 01 to an action. This part ships the binding infrastructure, the dispatcher, and the simplest action: send a fixed piece of text to the agent.

## Why it works at all

Short-click sequences already reach the app. `RingSync` calls `handleButtonPressSequence(...)` *above* the `if (transferStatus.isAudio)` branch, so a click-only collection still delivers its sequence even though the accompanying 0.001s of audio is discarded by the existing 1.0s duration floor. Nothing consumed those sequences except media control.

## Design notes for review

**Media control wins.** A binding on a click count media control owns would never fire, because the media map is consulted first. Rather than let users create dead config, `MusicControlMode.reservedClickCounts()` is the single source of truth: the dispatcher skips those counts and the settings UI locks them with an explanation. The precedence rule is extracted into a pure `resolveButtonDispatch()` so it can be unit-tested without the `expect` media calls.

**Dispatch happens off the sequence collector.** An action can take seconds (agent inference). Awaiting it inside `sequenceEvents.collect { }` would delay or drop the next gesture — including media control, which is currently instant. It's launched on `RecordingBackgroundScope` instead.

**Undecodable rows degrade, they don't throw.** The action is stored as JSON. A row this build can't parse becomes `ClickAction.Unsupported` rather than failing the query, matching the existing tolerant handling for `SemanticResult` and `RecordingDocument`.

## Schema

New `ClickActionBinding` table, v32 → v33 auto-migration, `clickCount` uniquely indexed. `33.json` is committed.

## Testing

- 202 unit tests pass (`:experimental:testAndroidHostTest`), 20 of them new: `ButtonDispatchTest` (precedence), `ClickActionRepositoryTest` (uniqueness, enable/disable, constraint races), `ClickActionCountsTest` (which counts are offered and why).
- Verified on a Pixel 10 (Android 17) with a real Index 01 (firmware 3.75.0), running **this branch alone**. Four physical clicks:

```
20:22:42.337  Undebounced: [Short]
20:22:42.419  Undebounced: [Short, Short, Short, Short]
20:22:43.122  Debounced:   [Short, Short, Short, Short]
20:22:43.133  Handled click action for 4 click(s): AgentText(...)
20:22:43.135  RecordingProcessingQueue: Handling text recording
20:22:44.944  CreateNoteTool: Creating note with text length: 31
```

- The `Unsupported` fallback was exercised against real data rather than only in a unit test: a row written by a later branch (`{"type":"tool_call",...}`) was read by this branch, logged a warning, rendered as "Unreadable — open to replace it", and did not crash. Reading it back on the later branch showed the row intact, so degrading the read does not rewrite the row.

- Not built for iOS — no macOS available. All code is in `commonMain` with no new `expect`/`actual` and no platform APIs, so I expect it to compile, but I haven't proven it and would appreciate a check.

## Screenshots

1-settings-row
2-bindings-list
3-editor-locked-counts

## AI assistance

This change was written with AI assistance. I have reviewed the implementation, understand it, and tested it on hardware.

### DIFF
--- a/experimental/schemas/coredevices.ring.database.room.RingDatabase/33.json
+++ b/experimental/schemas/coredevices.ring.database.room.RingDatabase/33.json
@@ -1,0 +1,1319 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 33,
+    "identityHash": "6576054b5f2741a416441285f88efec4",
+    "entities": [
+      {
+        "tableName": "LocalReminderData",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `time` INTEGER, `message` TEXT NOT NULL, `recordingId` TEXT DEFAULT NULL, `notifyBeforeMillis` INTEGER DEFAULT NULL)",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "time",
+            "columnName": "time",
+            "affinity": "INTEGER"
+          },
+          {
+            "fieldPath": "message",
+            "columnName": "message",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "recordingId",
+            "columnName": "recordingId",
+            "affinity": "TEXT",
+            "defaultValue": "NULL"
+          },
+          {
+            "fieldPath": "notifyBeforeMillis",
+            "columnName": "notifyBeforeMillis",
+            "affinity": "INTEGER",
+            "defaultValue": "NULL"
+          }
+        ],
+        "primaryKey": {
+          "autoGenerate": true,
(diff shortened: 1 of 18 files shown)


vishal-android-freak:
Follow-ups #328 and #329 are open as drafts and stack on this one, so the series can be read as a whole. They're drafts because until this merges GitHub shows their diffs cumulatively — each body says which commit to review.

ericmigi:
can you make these draft/cancel them for now? i'm working on a solution that will come out in a week or two that will solve for most of the needs you have here. if anything is still remaining after that, please open a new issue (not pr)

vishal-android-freak:
Sounds good. Will close these :)

PR #330, opened 2026-08-08
Refresh relative timestamps when Index resumes

jmole-turtlebot:
## Summary
- Refresh Index timestamps immediately when the app returns to the foreground.
- Update relative labels every minute while the screen is composed.
- Share the current time across recording, sync, and reminder formatters to ensure consistent recomposition.


jmole:
Hi Eric, I noticed that I was seeing stale timestamps each time I opened the app on iOS, like the latest note would always say "Just Now".

This patch should fix the issue.

PR #337, opened 2026-08-11
Save watch logs to the phone once a day

adipascu:
The watch only keeps a couple of days of logs, and getting them off currently needs a person: submitting a bug report, or sharing the report zip from the Bug Report screen. The pipelines that run on their own carry crash dumps and numeric metrics, not logs. So a firmware contributor collecting log data over a longer stretch, without access to the server side, has to remember to pull logs every couple of days or silently lose the oldest ones.

This adds an Auto-save watch logs toggle to the debug settings. When it is on, the daily background sync also gathers the connected watch's logs and keeps the newest 21 copies in the phone's Downloads folder under `Pebble watch logs`. If no watch is reachable it tries again on the next sync instead of skipping the day. Android 10 and newer write through `MediaStore`. Older Android versions and iOS fall back to app storage.

Off by default and hidden unless debug options are on, so nothing changes for regular users.

The other commit fixes the share sheet labeling the bug report zip as an audio file. `share()` was hardcoded to `audio/wav` from when recordings were its only payload, and now takes the type from the caller.

---

### DIFF
--- a/composeApp/src/commonMain/kotlin/coredevices/coreapp/ui/screens/BugReportScreen.kt
+++ b/composeApp/src/commonMain/kotlin/coredevices/coreapp/ui/screens/BugReportScreen.kt
@@ -280,7 +280,7 @@ fun BugReportScreen(
 
                         BugReportState.UploadingAttachments -> Unit
                         is BugReportState.ReadyToShare -> {
-                            platformShareLauncher.share(it.name, it.file)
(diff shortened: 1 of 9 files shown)


ericmigi:
This button already does this

adipascu:
You're right, dropped it. What's left is the daily auto-save, so logs outlive the watch's small buffer without manual pulls, and a fix for the zip being shared as an audio file.

PR #351, opened 2026-08-16
Let the user choose which phone calendar Index adds events to

### LINKED ISSUE #350: user can't choose or see which calendar index integrates with
Index "Integrations" offers to integrate with my "Phone calendar". But I have several: samsung calendar (long since disabled), three google calendars, and nextcloud calendar. Which one is it using?

I don't get to select, and there's no way to tell. Events are successfully created according to the timeline, but I can't find them anywhere.

After much searching, I re-enabled the samsung calendar app, and found a hidden "phone calendar" in its' settings (separate from the samsung calendar which I knew I wasn't using).

The integration UI should let the user choose and see, which calendar they are integrating.

ohthehugemanatee:
## Problem

Closes #350 

Connecting the Phone Calendar integration only grants calendar permission and flips a boolean. Events go to whatever calendar the platform picked, with no indication to the user. Most users have many calendars, and there's no guarantee the phone is offering Pebble the "right" one.

## Solution

This PR changes the UI for calendar integration, so the user can select from all known calendars. It also changes display to show which calendar is active.

## Changes

- `SystemCalendar.getWritableCalendars()`: lists calendars the user can create events in, platform default first. Android filters on `CALENDAR_ACCESS_LEVEL >= CAL_ACCESS_CONTRIBUTOR` and puts the account's own calendar first using the existing `findPrimaryWritableCalendarId`. iOS filters on `allowsContentModifications` and puts `defaultCalendarForNewEvents` first.
- `NewCalendarEvent.calendarId`: the calendar to write to. `createEvent` uses it when it's still writable and falls back to the default when it's null or the calendar has gone away (account removed), otherwise a stale selection would silently break event creation.
- Ring `Preferences.phoneCalendarId` stores the choice; `CreateCalendarEventTool` passes it through.
- `PhoneCalendarPicker`, a radio list of writable calendars with their account, shown when connecting the integration (Accounts → Add integration → Phone Calendar) and in its manage dialog. It defaults to the platform default, which is the first entry. 

Both platforms are covered; the Android provider query moved to `Dispatchers.IO` since the settings UI now calls it.

## Testing

- CI green (`android` + `ios`): `assembleDebug testDebugUnitTest testAndroidHostTest jvmTest lintDebug`.
- `WritableCalendarTargetTest` (commonTest) covers target resolution: requested calendar, no choice, calendar gone, no writable calendars.
- `OwnAccountFirstTest` (androidHostTest) covers the Android primary-calendar ordering, including that a row with no account name isn't treated as the account's own.

## manual testing

Not done. I'm a very seasoned developer professionally but not for mobile; i don't have any kind of real build environment set up outside of the included CI. I also have no access to an iOS device. Since this change is so small and architecturally straightforward, I felt comfortable reviewing and  contributing it. 

This is entirely written through claude, with separate reviews both by me and by copilot.

### DIFF
--- a/androidApp/src/androidTest/kotlin/coredevices/coreapp/evals/RingRecordingE2ETest.kt
+++ b/androidApp/src/androidTest/kotlin/coredevices/coreapp/evals/RingRecordingE2ETest.kt
@@ -702,6 +702,7 @@ private class E2EPreferences : Preferences {
     override val autoDismissActionNotifications: StateFlow<Boolean> = MutableStateFlow(true)
     override val backupEnabled: StateFlow<Boolean> = MutableStateFlow(false)
     override val phoneCalendarEnabled: StateFlow<Boolean> = MutableStateFlow(false)
+    override val phoneCalendarId: StateFlow<String?> = MutableStateFlow(null)
     override val useEncryption: StateFlow<Boolean> = MutableStateFlow(false)
     override val encryptionKeyFingerprint: StateFlow<String?> = MutableStateFlow(null)
     override val lastWipedRing: StateFlow<String?> = MutableStateFlow(null)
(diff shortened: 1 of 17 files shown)


ohthehugemanatee:
bump @crc-32  any chance for a review?

ChristophDerndorfer:
This continues to be the single biggest feature gap in the day-to-day use of the Index01.

PR #378, opened 2026-08-25
Deliver third-party data logging to companion apps over PebbleKit 2

neelts:
Data logging sessions of third-party watchapps are dropped after the protocol-level ACK, so a watchapp cannot get its spooled data to a companion app (#160, #201). #292 tried this with the legacy Intent broadcasts and was closed in favor of PebbleKit 2. The API side landed in [pebble-dev/PebbleKitAndroid2#37](https://github.com/pebble-dev/PebbleKitAndroid2/issues/37) and is released in io.rebble.pebblekit2:server 1.3.0.

Datalogging now emits third-party sessions as one ordered flow, thirdPartyEvents (Batch / Finished), instead of dropping them. Each event carries the session identity (UUID, tag, open timestamp), the item size and the watch serial.

On Android, PebbleKit2Datalogging collects the flow and calls the new sendOnDataLogReceived / sendOnDataLogSessionFinished. 
One worker per session sends the batches in sequence, and sends the finish event after the companion app acknowledged all batches. 
Nack, Unknown, null, a binder failure or a call timeout causes a limited number of retries; then the worker discards the rest of the session. 
Delivery is not tied to a running watchapp, because spooled data arrives on reconnect while no watchapp is open. 
The companion package comes from the watchapp appinfo.json (companionApp.android.apps[].package). 
The delivery state machine is covered by unit tests (DatalogDeliveryTest).

I validated the flow end to end with an Intent-based prototype of this change (a fully offline session, 114/114 records received, zero loss), I will repeat the test on this PebbleKit 2 path and report here.

cc @sjp4 this is the PebbleKit 2 successor of #292


libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/two/PebbleKit2Datalogging.kt
```diff
@@ -0,0 +1,78 @@
...
+                    WatchIdentifier(event.watchSerial),
+                )
+                return result == ReceiveResult.Ack
+            }
+
+            override fun close() = connector.close()
+        }
+    }
+
+    private suspend fun companionPackagesFor(uuid: Uuid): List<String> {
```
sjp4: Might want to put this on Dispatchers.IO

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/two/DatalogDelivery.kt
```diff
@@ -0,0 +1,167 @@
...
+            sessions.remove(key)
+            channel.close()
+        }
+        return null
+    }
+
+    private suspend fun deliverWithRetry(send: suspend () -> Boolean): Boolean {
+        repeat(attempts) { attempt ->
+            val stored = try {
+                withTimeoutOrNull(sendTimeout) { send() } == true
```
sjp4, line 140: Looks like PK2 doesn't support handling cancellation if this times out (so it will just get stuck here) - may be worth addressing in PK2

sjp4 (commented):
Did you test with PK2? (PR description says you are going to, doesn't say whether you did)

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/two/DatalogDelivery.kt
neelts: thanks! something like this?
https://github.com/pebble-dev/PebbleKitAndroid2/pull/51

libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/pebblekit/two/PebbleKit2Datalogging.kt
neelts: Applied the fix now

neelts:
> Did you test with PK2? (PR description says you are going to, doesn't say whether you did)

Not yet, only with the tests, will try this week

neelts:
Actually managed to test just now! All works end to end :tada:

PR #388, opened 2026-08-29
Adjust image color quantization to better match colors and use a more realistic palette

aveao:
The new notification pictures feature is neat but the pictures appear pretty low-contrast, and not all colors match the on-screen color well.

I tried quiite a few different things, which I'll document in a separate issue, but one of them was improving the color matching logic to consider distance to the palette considering all channels, rather than taking the distance on each channel individually.

Other is perhaps more controversial, which is swapping the palette for comparison to one I measured a while back. This involved taking dozens of pictures of a Time 2 (and PTS but that's unrelated) under various lighting conditions with a DSLR, next to a color calibration grid (SpyderCheckr iirc), and then applying correction in darktable, before sampling the color grid off the device. It's a bit tricky to dial in, especially considering reflective screen and variability of contrast, but I made this out of it: https://fgl.ave.zone/pebblecolors.html#pt2sunlight

It might benefit from a separate measurement for time round, which I'll do when I get mine (assuming there's sunshine hours left in the year).

---

On device (also with Bayer 8x8 in this case):

PXL_20260829_100345224

---

Written with Claude Opus 5 but I can understand & hopefully explain the logic :) I'd normally clean up the code some more but the previous state already had claudisms so I kept the new claudisms added too, let me know if you prefer it cleaned.

---

Demo: https://fgl.ave.zone/image-recolor-demo/

Some screenshots from it (note that these are emulating the actual screen colors, try toggling bottom left switch):

Top left is current, bottom right is proposed. Do note that the colors are emulating my measured screen values, if using the demo tool yourself you can toggle that on bottom left.

Screenshot 2026-08-29 at 17-34-00 Palette _ matching comparison
Screenshot 2026-08-29 at 17-33-43 Palette _ matching comparison
Screenshot 2026-08-29 at 17-33-34 Palette _ matching comparison
Screenshot 2026-08-29 at 17-33-21 Palette _ matching comparison
Screenshot 2026-08-29 at 17-32-41 Palette _ matching comparison

### DIFF
--- a/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/imaging/ImageEncoder.kt
+++ b/libpebble3/src/commonMain/kotlin/io/rebble/libpebblecommon/imaging/ImageEncoder.kt
@@ -2,14 +2,15 @@ package io.rebble.libpebblecommon.imaging
 
 /**
  * Encodes an ARGB image as the watch's 4-bpp palettized [EncodedImage]: choose a 16-colour palette
- * by median cut over the watch's 64-colour (GColor8) space, Floyd–Steinberg dither to that palette,
- * and pack two 4-bit indices per byte (even x = high nibble, matching the firmware).
+ * by median cut over the watch's 64 screen colors, Floyd–Steinberg dither to that palette, and
+ * pack two 4-bit indices per byte (even x = high nibble, matching the firmware).
  *
  * Transparency is binary, and costs a palette slot when present. The watch only skips these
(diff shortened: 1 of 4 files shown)


aveao:
Undrafted after testing for a bit.

I had some images simply not show so I panicked and drafted, but I think that was due to unrelated issues, I've had several looks over the code and it's not due to transparency issues afaict. Almost all images arrive and render fine.

Generally it'd be cool to have image support in the notification test UI.

PR #421, opened 2026-09-12
Add scroll hint to notification filter controls

OPerepadia:
In Notifications, the "Show system apps" filter was hidden off-screen on narrow displays, with no indication that the filter row could be scrolled horizontally.

This change adds a scroll hint to the filter controls, using same animation as on Apps page

Tested on Android with a 360 dp-wide display.

Assisted-by: Codex:gpt-5


sjp4:
I'd prefer an indication that you can scroll, rather than wrapping. On the apps page we have an auto-scroll the first time you see it to indicate that it can scroll. Open to other UX ideas to have the same effect

OPerepadia:
Yes I agree, this is cleaner. I replaced wrapping with the same scroll hint as on the apps page.

PR #423, opened 2026-09-16
index: new items follow their list's kind, add Switch to checklist menu entry

### LINKED ISSUE #416: Index: let a user-created list default to checklist items, like Shopping list does
## Summary

The seeded **Shopping list** turns every new item into a tickable checklist item, but a list the user creates always produces plain note items, and there is no setting to change that. I'd like a per-list "default item type" so a new list (camping, packing, hardware store) can behave like Shopping.

## Prior reports

This has come up repeatedly on the forum, and a Core staff member has already acknowledged it as a bug via an in-app report (no GitHub issue exists yet, which is why I'm opening this one to track it):

- [How to add Checklist Items to a List?](https://forum.repebble.com/t/how-to-add-checklist-items-to-a-list/1355) (2026-08-13). Trevor replied: "Glad to hear you submitted your bug report. This is not the intended behavior and will be fixed."
- [Automate Checklists](https://forum.repebble.com/t/automate-checklists/1343) (2026-08-12)
- [Managing Notes is so tedious right now](https://forum.repebble.com/t/managing-notes-is-so-tedious-right-now/1280) (2026-08-04), which asks for a per-list default type explicitly
- [Default to 'checklist' type for all index notes](https://forum.repebble.com/t/default-to-checklist-type-for-all-index-notes/1490) (2026-08-29)
- [New checkbox list](https://forum.repebble.com/t/new-checkbox-list/1580) (2026-09-07)

## What happens today

- Type an item into a user-created list: it is a note (dash), not a checklist item (circle).
- Dictate "add batteries to my camping list" with the ring: same, the item lands in the right list but as a note.
- Only Shopping gets checklist items. The list's overflow menu offers "Rename list" and "Delete list" only.

## Where it comes from (as of `40ea15c8`)

The data model already has the setting. `ListDocument.listKind` (`index-ai/.../data/entity/ListDocument.kt:33`) is `"note" | "checklist" | "bullets"`, mirrored into Room and synced to Firestore, and Shopping is seeded with `listKind = "checklist"`. But nothing lets a user set it (`ListRepository.createList` always defaults to `"note"`), and both item-creation paths ignore `listKind` and special-case Shopping by id:

- Typed items: `ObjectDetailViewModel.createChildItem` (`ObjectDetailViewModel.kt:421`) → `if (s.list.firestoreId == LIST_SHOPPING_ID) "checklist" else kind`
- Ring dictation: `ItemFactory.noteItem` (`ItemFactory.kt:100`) → `if (parentId == LIST_SHOPPING_ID) ItemMetadata.Checklist else ItemMetadata.Note`

Rendering is already per-item (`EditableNoteRow` checks `child.kind == "checklist"`), so mixed lists display correctly.

## Proposal

- Add a "Switch to checklist" / "Switch to notes" entry to the list's overflow menu, next to "Rename list". Shown for every list except Reminders.
- Make both creation paths read the list's `listKind` instead of checking for the Shopping id. Shopping keeps working because it is seeded as `"checklist"`.
- Leave existing items untouched when a list is switched. Only new items pick up the new default.
- Out of scope: converting existing items, exposing the unused `"bullets"` kind, changing the "Add note" placeholder.

No DB migration needed. The change is about 140 lines across 3 source files and 4 test files, all in `experimental/`.

## Status

I have a working branch for this, tested on a Pixel 9 Pro (Android 16) with an Index 01 against a personal Firebase project: typed and dictated items into a checklist-kind list get a tick circle, flipping the list leaves old items alone, Shopping and Notes to self behave as before. The change is entirely in `commonMain`, so it covers iOS as well, but I only have Android hardware and haven't run it on iOS. Happy to open a PR if this is something you'd take. AI assistance was used in writing it and I've reviewed the diff.

redparchel:
Fixes #416.

## What

Only the seeded Shopping list produces checklist items. Both item-creation paths special-case it by id (from MOB-8946), and although `ListDocument.listKind` already exists and is synced, nothing lets a user set it on a list they created. So every user-made list gets plain notes with no way to change that.

Three small changes, all in `experimental/` `commonMain`:

1. **`ItemFactory.noteItem`** (ring dictation, share, `create_list_item`) reads the parent list's `listKind` through `ListRepository` instead of comparing against `LIST_SHOPPING_ID`. `noteItem` becomes `suspend`; all callers already are. Koin resolves the new constructor parameter since `ListRepository` is already a singleton.
2. **`ObjectDetailViewModel.createChildItem`** (typed items) does the same, and a new `setListKind` writes the list document back with the chosen kind, modelled on `renameList`.
3. **`ObjectListDetail`** overflow menu gains "Switch to checklist" / "Switch to notes" next to "Rename list", for every list except Reminders.

Existing items are untouched when a list is switched; only new items follow the new default. Shopping behaves exactly as before because it is seeded with `listKind = "checklist"`. No migration: the field, its Room column and Firestore sync all predate this change. The unused `"bullets"` kind is deliberately not exposed.

## Verification

- `:experimental:testAndroidHostTest` passes (311 tests). Existing Shopping tests still pass; new cases cover a checklist-kind custom list, a note-kind custom list, and the `create_list_item` and `create_note` paths into a checklist-kind list.
- `:androidApp:assembleDebug` and `lintDebug` pass.
- On a Pixel 9 Pro (Android 16) with an Index 01: typed items into a switched list get the tick circle, dictated items into a checklist-kind list arrive as checklist items, switching back leaves old items as they were, Shopping and Notes to self unchanged.
- Not run on iOS (no device); the change is `commonMain` only.

## AI disclosure

Per CONTRIBUTING.md: written with AI assistance (Claude). I directed the design (menu toggle, existing items left alone), tested it on hardware, and reviewed the diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crc-32:
Hey! Apologies, this has been recently implemented by us and will be in a future release. Thanks for the contribution though!

redparchel:
Awesome! I'm just happy this concept is making it in!

